PR #1, opened 2024-07-29
Minor fixes: `sqrt()` and `!=`

yutannihilation:
Just happened to find these errors. I know this is under very active development, so please feel free to ignore!

`sqrt()`:

```
external function sqrt cannot be found
```

`!=`:
```
   │
 7 │   if (phasor(freq) != 0.0) {
   ·                    ─┬
   ·                     ╰── Variable ne not found in this scope
```


tomoyanonymous:
Thanks!!!

PR #4, opened 2024-08-03
Parse `^`

yutannihilation:
It seems the lexer has a definition for `^`, but it's missing from the regex pattern and bytecodegen.

https://github.com/tomoyanonymous/mimium-rs/blob/fadac989af8bc2a54a699080cbc5bf53fec223bc/mimium-lang/src/compiler/parser/lexer.rs#L66

### DIFF
--- a/mimium-lang/src/compiler/bytecodegen.rs
+++ b/mimium-lang/src/compiler/bytecodegen.rs
@@ -493,6 +493,7 @@ impl ByteCodeGenerator {
             mir::Instruction::MulF(v1, v2) => self.emit_binop2(VmInstruction::MulF, &dst, v1, v2),
             mir::Instruction::DivF(v1, v2) => self.emit_binop2(VmInstruction::DivF, &dst, v1, v2),
             mir::Instruction::ModF(v1, v2) => self.emit_binop2(VmInstruction::ModF, &dst, v1, v2),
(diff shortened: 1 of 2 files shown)


tomoyanonymous:
Thanks, I'll add some tests later.

yutannihilation:
Thanks! I'm yet to figure out what test should be where. I'll learn from the upcoming commits.

tomoyanonymous:
yeah, briefly explainied here, there is `integration_test.rs` in the `mimium-lang/tests` directory.

The simplest example is `adder` one.

https://github.com/tomoyanonymous/mimium-rs/blob/ee0ed1a0cc5c3c8902bf95990d3209f2645e888d/mimium-lang/tests/intergration_test.rs#L25-L30

https://github.com/tomoyanonymous/mimium-rs/blob/ee0ed1a0cc5c3c8902bf95990d3209f2645e888d/mimium-lang/tests/mmm/adder.mmm#L1-L6

You can test the value of the return value of `dsp` function of the file located in `tests/mmm`  like `run_file_test("adder.mmm", 3)`. The second argument is the how many time samples you inspect the audio stream.
So, pure function can be simply tested with like `run_file_test("modf.mmm", 1)` and compare result with vector of 1 element.

yutannihilation:
I see. What I was wondering was to add such an integration test per operator, or some unit test to `compiler/parser/tests.rs`.
But, apparently, the integration tests should cover arithmetic operators. I'll try later. Thanks for the explanation!

PR #6, opened 2024-08-04
chore: Rmove `then` token

yutannihilation:
While `Expr` has `Then` to represent the then branch, it seems `Token` doesn't need `Then` because mimium's `if` statement doesn't have `then` keyword.

https://mimium.org/en/docs/users-guide/language-specification/grammar-rule/


tomoyanonymous:
Indeed... Maybe I have put it when I was re-designing language spec but no longer needed for the token.

PR #13, opened 2024-08-04
Add more integration tests for simple arithmetic operations

yutannihilation:
This is a followup to https://github.com/tomoyanonymous/mimium-rs/pull/5; for simple test cases, probably it's better to write the expression directly rather than creating a file per case.

Note that, the current code runs as a single test case and stops immediately on the first failure, so this can be a bit tiresome to repeat running and fixing when this gets more cases. In that case, some framework like [rstest](https://github.com/la10736/rstest) might be helpful (I haven't used this...).

### DIFF
--- a/mimium-lang/src/utils/error.rs
+++ b/mimium-lang/src/utils/error.rs
@@ -14,8 +14,11 @@ pub trait ReportableError: std::error::Error {
     }
 }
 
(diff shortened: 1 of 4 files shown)


tomoyanonymous:
Looks smart! These tests are mostly for checking regression, so it is ok to do it sequentially for now.

PR #14, opened 2024-08-07
Fix fncall

yutannihilation:
As you probably already notice, the current `dev` branch fails to process `statefn2_same.mmm`. This pull request is an attempt to fix this by partly reverting the change introduced in #3.

```
thread 'cpal_wasapi_out' panicked at library\core\src\panicking.rs:221:5:
unsafe precondition(s) violated: slice::get_unchecked requires that the index is within the slice
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread caused non-unwinding panic. aborting.
error: process didn't exit successfully: `target\debug\mimium-cli.exe mimium-lang/tests/mmm/statefn2_same.mmm` (exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN)
```

The stack trace indicates the panic occurs here:

https://github.com/tomoyanonymous/mimium-rs/blob/f2bb45ec4d4c46ffc05d4b54f72587005ea35a9f/mimium-lang/src/runtime/vm.rs#L36

It seems `rawdata` is resized by the function's `state_size`.

https://github.com/tomoyanonymous/mimium-rs/blob/f2bb45ec4d4c46ffc05d4b54f72587005ea35a9f/mimium-lang/src/runtime/vm.rs#L625

The `state_size` is the sum of `statesize` supplied from the caller of `emit_fncall()`.

https://github.com/tomoyanonymous/mimium-rs/blob/f2bb45ec4d4c46ffc05d4b54f72587005ea35a9f/mimium-lang/src/compiler/mirgen.rs#L272-L274

So, my guess is `Instruction::PushStateOffset()` and `push_sum` should use `statesize` instead of the offset from the context. But, honestly I'm not confident if my understanding is correct...

### DIFF
--- a/mimium-lang/src/compiler/mirgen.rs
+++ b/mimium-lang/src/compiler/mirgen.rs
@@ -275,13 +275,12 @@ impl Context {
             self.push_inst(Instruction::Uinteger(idx))
         };
         //insert pushstateoffset
-        let coffset =  self.get_ctxdata().state_offset;
-        if coffset > 0 {
+        if self.get_ctxdata().state_offset > 0 {
             self.get_current_basicblock().0.push((
                 Arc::new(Value::None),
-                Instruction::PushStateOffset(coffset),
+                Instruction::PushStateOffset(statesize),
             ));
-            self.get_ctxdata().push_sum += coffset;
+            self.get_ctxdata().push_sum += statesize;
         }
 
         let res = self.push_inst(Instruction::Call(f.clone(), args));


tomoyanonymous:
Hmm, I have not noticed this because `statefn2_same.mmm` passes the test on my environments(M1 Mac & WSL2). Maybe the check of unsafe precondition is more strict on Windows?
Anyway, this partial revert works without problem on my environment.
This implementation of internal state size calclation is still complicated for me, we need to figure out more robust implmentation without using destructive assignment directly

yutannihilation:
Oh, you didn't notice this? (I thought this is the reason why `dev` branch is not merged into `main` yet.)

> Maybe the check of unsafe precondition is more strict on Windows?

Sorry, I didn't describe the details. This check is an additional one which is enabled only on debug build. I found this error on `cargo test`, but this doesn't show up if I run the file with `cargo run --release`.

> This implementation of internal state size calclation is still complicated for me, we need to figure out more robust implmentation without using destructive assignment directly

Thanks. I hope we can figure it out soon!

Issue #12, opened 2024-08-04
Use  `id_arena` and `InternedString` in the compiler


tomoyanonymous:
In the current implementation, compiler clones many `String` value and comparing them.
Use `StringInterner` crate will improve the performance.

https://docs.rs/string-interner/latest/string_interner/

Also, recursive type definitions in the `Expr` and `Type` are realized simply with `Box<Self>`, however, this this will make a implementation
of interactive inspection of values difficult, for example, for the language server. Using `id_arena` simplifies this situation, along with performance improvements.

They are tested on my another project otopoiesis (an experimental branch)

https://github.com/tomoyanonymous/otopoiesis/blob/372bfb7b5c760744f69532c63e8804878a84397d/otopoiesis_lang/src/expr.rs#L31-L32

https://github.com/tomoyanonymous/otopoiesis/blob/372bfb7b5c760744f69532c63e8804878a84397d/otopoiesis_lang/src/lib.rs#L25-L28

PR #15, opened 2024-08-07
Add Test CI

### LINKED ISSUE #7: Add Test CI

yutannihilation:
Close #7

Some notes:

* Currently this fails because of the problem I reported in #14.
* The config matrix contains only Ubuntu for now because no difference is found so far between platforms. But, it's easy to add macOS and Windows as well.
* The setup for ALSA is derived from [cpal's GitHub Actions settings](https://github.com/RustAudio/cpal/blob/master/.github/workflows/cpal.yml).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,34 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [main, dev, ci]
+  push:
+    branches: [main, dev, ci]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  cargo_test:
+    name: "cargo test (OS: ${{ matrix.config.os }}, toolchain: ${{ matrix.config.toolchain }}"
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - { os: ubuntu-latest,  toolchain: 'stable'  }
+          - { os: ubuntu-latest,  toolchain: 'nightly' }
+          # - { os: macOS-latest,   toolchain: 'stable'  }
+          # - { os: windows-latest, toolchain: 'stable'  }
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@master
+        with:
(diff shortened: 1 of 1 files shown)


yutannihilation:
You can see how this CI setup runs (and fails) here:

https://github.com/yutannihilation/mimium-rs/actions

tomoyanonymous:
Thanks! This really helps.

In the config, both `stable` and `nightly` channels are tested but `stable` channel looks like being overwritten to nightly by `rust-toolchain.toml`

https://github.com/yutannihilation/mimium-rs/actions/runs/10293042761/job/28488605866#step:3:67

```
Run rustup default stable
info: using existing install for 'stable-x86_64-unknown-linux-gnu'
info: default toolchain set to 'stable-x86_64-unknown-linux-gnu'

  stable-x86_64-unknown-linux-gnu unchanged - rustc 1.80.0 (0514789[57](https://github.com/yutannihilation/mimium-rs/actions/runs/10293042761/job/28488605866#step:3:64) 2024-07-21)
info: note that the toolchain 'nightly-x86_64-unknown-linux-gnu' is currently in use (overridden by '/home/runner/work/mimium-rs/mimium-rs/rust-toolchain.toml')
```

I set nightly as a default because I use some unstable feature, especially `box_pattern` and `iterator_try_collect` to simplify the codes. So, I think we need to test only nightly toolchain for now.

`box_pattern` seems to hardly be stabilized, however, if we start using arena for recursive types(#12), `box_pattern` will no longer needed anyway.

https://github.com/rust-lang/rust/issues/29641

yutannihilation:
Oh, I see. I forgot `rust-toolchain.toml` exists. Then I remove the `stable` case.

Thanks for explaining the context including #12. Since this project is mainly about providing a binary, not a library, I guess compatibility with stable is not very important. But, #12 looks nice. I'll try to understand.

yutannihilation:
result:
https://github.com/yutannihilation/mimium-rs/actions/runs/10295092683

(mainly note to self)
I quickly searched if there's some way to let `dtolnay/rust-toolchain` refer to `rust-toolchain.toml` instead of specifying `toolchain: nightly`, but it seems the author doesn't want to support it. So, probably there's no nicer way.

https://github.com/dtolnay/rust-toolchain/issues/77

PR #17, opened 2024-08-08
Implement unary `-`

yutannihilation:
Currently, this code

```rs
fn dsp() {
    2.0-2.0
}
```

cannot be parsed.

```
Filename: \\?\C:\Users\Yutani\Documents\GitHub\mimium-rs\tmp.mmm
, ror: unexpected token something else, expected ^, ;,
, (, |, -, /, ", :, *, <, !, =, 0, ., +, }, [, ), %, ,, &, {, ], >
   ╭─[\\?\C:\Users\Yutani\Documents\GitHub\mimium-rs\tmp.mmm:2:3]
   │
 2 │     2.0-2.0
   ·   ┬
   ·   ╰── Unexpected token
───╯
Error: unexpected token while parsing apply, expected ||, linebreak, }, |>, >, -, <=, +, !=, /, ;, >=, &&, %, (, ==, <, *, ^
   ╭─[\\?\C:\Users\Yutani\Documents\GitHub\mimium-rs\tmp.mmm:2:8]
   │
 2 │     2.0-2.0
   ·        ──┬─
   ·          ╰─── Unexpected token -2.0
───╯
```

This is because there's no parser for unary operations. Currently, the lexer parses `-` as part of a float literal.

https://github.com/tomoyanonymous/mimium-rs/blob/ee0ed1a0cc5c3c8902bf95990d3209f2645e888d/mimium-lang/src/compiler/parser/lexer.rs#L24-L30

It might be a choice to require spaces before and after `-` to distinguish the binary op and the unary op. But, I think users expect this syntax is valid, so mimium anyway needs a unary parser.

```rs
-x
```

### DIFF
--- a/mimium-lang/src/ast_interpreter.rs
+++ b/mimium-lang/src/ast_interpreter.rs
@@ -83,13 +83,14 @@ impl Context {
     }
 }
 
-const EXTERN_ENV: [&str; 27] = [
-    "add", "sub", "mult", "div", "mod", "eq", "ne", "le", "lt", "ge", "gt", "atan2", "sin", "cos",
-    "not", "round", "floor", "ceil", "atan", "sqrt", "abs", "min", "max", "pow", "log", "print",
-    "println",
+const EXTERN_ENV: [&str; 28] = [
+    "neg", "add", "sub", "mult", "div", "mod", "eq", "ne", "le", "lt", "ge", "gt", "atan2", "sin",
+    "cos", "not", "round", "floor", "ceil", "atan", "sqrt", "abs", "min", "max", "pow", "log",
+    "print", "println",
 ];
 
 fn lookup_extern_env(name: &str) -> Option<&str> {
+    println!("{name}");
     let filtered = EXTERN_ENV
         .into_iter()
         .filter(|n| *n == name)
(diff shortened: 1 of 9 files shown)


tomoyanonymous:
Thanks! Maybe should we add more test cases with nested operation to ensure the parse works correctly?
```rust
    run_simple_test("- -1.0", 1.0, 3);
    run_simple_test("-(-hoge)", 2.0, 3);
```

yutannihilation:
Thanks. Sure, I'll add some more test cases.

yutannihilation:
It seems a function call like this is not handled properly, so my implementation is not correct. Let me think.

```rs
-cos(0.0)
```

tomoyanonymous:
Now parser emits ast for `-cos(0.0)` to `(app (app neg (cos))((float 0.0))))`. (`(- cos) (0.0)`)
(You can emit raw ast as lisp-like syntax with `--emit-ast` option )

Now precedence of `unary` operator is bigger than `apply`. It should be just in between `apply` and other binary operators.

yutannihilation:
> Now precedence of `unary` operator is bigger than `apply`. It should be just in between `apply` and other binary operators.

You are right, thanks. Should be fixed now!

Issue #8, opened 2024-08-04
Implement Tuple(Array?) Type


tomoyanonymous:
For stereo and the more multi-channel processing, tuple types needs to be implemented.

https://github.com/mimium-org/mimium/blob/dev/examples/demo.mmm

tomoyanonymous:
Goal: Implment biquad filter (direct form 2) like below.

```rust
fn biquad(x,coeffs){
    let (a,a1,a2,b1,b2) = coeffs;
    let inner = |x|{
		let (ws, wss) = self
		let w = x - a1*ws - a2*wss
		(w, mem(w))
	}
	let (w,ws) = inner(x);
    let wss = mem(ws);
	b0*w + b1*ws + b2*wss
}
```

PR #20, opened 2024-08-11
Add preliminary tuple support

yutannihilation:
~~This pull request addresses #8 only very partially. I was hoping I could implement generation of MIR and bytecode, but it seems I'm still far from it. So, let me open this pull request for now.~~

Update: this pull request now adds a barely-working support for tuple.

Now the code like this

```
fn to_stereo(x:float) -> (float,float) {
  (x, x)
}

fn dsp() {
    to_stereo(1.0)
}
```

can be parsed to the following AST.

```
(let
    _mimium_global
    (lambda
       ()
       (let
          to_stereo
          (lambda
             (
                (tid
                   x
                   number))
             (tuple
                (x
                   x)))
          (let
             dsp
             (lambda
                ()
                (app
                   to_stereo
                   (
                      (float
                         1.0))))
             )))
    )
```

### DIFF
--- a/mimium-audiodriver/src/backends/cpal.rs
+++ b/mimium-audiodriver/src/backends/cpal.rs
@@ -32,15 +32,15 @@ struct NativeAudioData {
     dsp_ochannels: usize,
     buffer: HeapCons<f64>,
     localbuffer: Vec<f64>,
-    count:u64
+    count: u64,
 }
 unsafe impl Send for NativeAudioData {}
 
 impl NativeAudioData {
     pub fn new(program: vm::Program, buffer: HeapCons<f64>) -> Self {
         let dsp_i = program.get_fun_index("dsp").expect("no dsp function found");
-        //todo! infer from program
-        let dsp_ochannels = 1;
+        let (_, dsp_func) = &program.global_fn_table[dsp_i];
+        let dsp_ochannels = dsp_func.nret;
 
         let vmdata = RuntimeData::new(program);
         let localbuffer: Vec<f64> = vec![0.0f64; 4096];
@@ -50,7 +50,7 @@ impl NativeAudioData {
             dsp_ochannels,
             buffer,
             localbuffer,
-            count:0
+            count: 0,
         }
     }
     pub fn process(&mut self, dst: &mut [f32], h_ochannels: usize) {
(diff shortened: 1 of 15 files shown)


yutannihilation:
Now tuple is supported to some extent in this pull request.

This code works. A sine wave can be heard only from the left channel.

```
let pi = 3.14159265359
let sr = 44100.0
fn phasor(freq){
  (self + freq/sr)%1.0
}

fn osc(freq){
  sin(phasor(freq)*pi*2.0)
}
fn dsp() -> (float,float) {
  (osc(440.0), 0.0)
}
```

However, this doesn't work. This compiles fine, but generates sine wave on both channels. I'm still figuring out what's wrong here.

```
...snip...

fn to_stereo(x:float) -> (float,float) {
  (x, 0.0)
}

fn dsp() -> (float,float) {
  to_stereo(osc(440.0))
}
```

tomoyanonymous:
This is great!! I'll check the details in this weekend but certainly we need to add nret to call operation (multiple value returns are partially implemented because I referred
the tutrial on the Lua for the bytecode design
as well but not sure).

yutannihilation:
Thanks! No hurry at all. Also thanks for your helpful advice. Then, I'll try what the tutorial does first.

What I'm wondering is, since Mimium is different from Lua in that it's a statically-typed language, there might be some shortcut on this process by utilizing type information somehow. Anyway, I believe this can be improved later and is not mandatory for now. I'll study first...

yutannihilation:
(No need to reply. Just a note.)

After adding `nret` to `Call`, it works now. I think I've done what I planned to do in this pull request, although this still has room for improvement.

One design I'm not fully confident with is the addition of a new instruction `MoveRange`. This could be implemented as a series of `Move`, but I guess it might be helpful to treat this as a single instruction in order to optimize in `remove_redundunt_mov()`, which I don't touch in this pull request. But, this is probably beyond my knowledge at the moment.

tomoyanonymous:
Ok, it looks working.
I was thinking to change Move to 3 arguments operator, like you made `MoveRange` (`Move` becomes just a alias to `MoveRange x y 1`).
I guess it might have some errors with the combination of let_tuple unpacking and variable escaping with closure, I will fix them in another PR.

yutannihilation:
> I was thinking to change Move to 3 arguments operator

I see. That also sounds good to me.


> I guess it might have some errors with the combination of let_tuple unpacking and variable escaping with closure, I will fix them in another PR.

Thanks for catching! I didn't figure out how to handle let_tuple in this pull request, and didn't notice about the problem with closure.

PR #21, opened 2024-08-15
nit: Pass colors from ColorGenerator

yutannihilation:
As [Ariadne's example code](https://docs.rs/ariadne/latest/ariadne/) shows, `ColorGenerator` can generate a color per error. Currently, this is not very useful, but this can look nice when the parser gets smarter. If you prefer the plain red and green (or any other colors), that's also fine to me.

**Before**

image

**After**

image

(Btw, it seems "Er" in the first red "Error" somehow disappears)

### DIFF
--- a/mimium-lang/src/compiler/parser/error.rs
+++ b/mimium-lang/src/compiler/parser/error.rs
@@ -35,7 +35,7 @@ where
     fn get_span(&self) -> Span {
         self.0.span()
     }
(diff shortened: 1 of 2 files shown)


tomoyanonymous:
Looks okay!  Though currently we don't have feature to collecting multiple errors except for the parser, but anyway, thanks.

PR #22, opened 2024-08-15
Implement tuple unpacking at let

tomoyanonymous:
~~closes~~ #8 
continued from #20.

It implements unpacking tuple at `let` like the code below.
```rust
fn gen_tup(){
    (5.0,2.0,3.0)
}
fn dsp(){
    let (a,b,c) = gen_tup()
    a+b*c
}
```

This PR adds `Pattern` and `TypedPattern` for the `Let` AST node by extending `TypedId`, rather than adding `LetTuple` node, for the nested tuple unpacking and maybe unpacking struct type values in the future.

In MIR, `GetElement` instruction has added to calculate offset of the virtual memory address for aggregate type members. This is mostly same as `getelementptr` in LLVM.


tomoyanonymous:
Remaining tasks: 
- [ ] add tests for nested tuple type unpacking
- [ ] add tests for the combination of tuple and escaping closure

tomoyanonymous:
(Memo) Currently some non-stateful codes returns different result between time = 0 and time > 0.

```rust
thread 'closure_open_3nested' panicked at mimium-lang/tests/intergration_test.rs:192:5:
assertion `left == right` failed
  left: [-3.0, 2.0]
 right: [2.0, 2.0]
```

tomoyanonymous:
last todo:
Virtual register mapping for non-primitive value(especially upvalue)

tomoyanonymous:
true final todo:

- [ ] appropriate type inference for `self` w/ non-primitive type

tomoyanonymous:
The code below cannot be compiled correctly yet because the type inference for `self` fails and stack allocation sizes goes wrong. However, it will be big change to the code to fix this type inference algorithm, so I'm going to merge this PR for now.

```rust
fn bifb()->(float,float){
    let (a,b) = self
    (a+1.0,b+2.0)
}

fn dsp(){
    bifb()
}
```

Quick Note:
- allocation size specification is added for bytecode instructions below
 - Move(MoveRange)
 - Set/GetUpValue
 - Set/GetGlobal (`GlobalPos` was changed from u16 to u8)
 - Set/GetState
- `Type::size` was moved to `BytecodeGenerator::word_size_for_type` because different backends may produce different sizes(though the function is static method currently)
- Copying from slice to slice in `vm.rs` contains many unsafe code now, and I tried efficient copying implementation with `slice::copy_from_slice(=memcpy)` and `slice::copy_within(=memmove)` but it produces some unexpected behaviour, so partly the copyings are implemented with iterator simply. There are many space for refactoring.

yutannihilation:
Great progress!!

By the way, aside from the issue with type inference, should `let TUPLE = NON_TUPLE` work? Currently, it seems the following code gives `(1.0, 0.0)`, but I believe this should be either `(1.0, 1.0)` or rejected because of the type mismatch. 

```
fn dsp(){
  let (x, y) = 1.0
  (x, y)
}
```

yutannihilation:
I see. Thanks for clarification.

PR #23, opened 2024-08-16
[PoC] Add "expression statement"

yutannihilation:
(This pull request is mainly to show the basic idea. Probably I'll abandon this and retry after #22 gets merged)

My current interest is to make this work.

```
fn dsp() {
  1.0
  1.0
}
```

At the moment, the parser fails to parse this.

```
Error: unexpected token while parsing function decl, expected linebreak, }, ;
   ╭─[path/to/tmp.mmm:3:3]
   │
 3 │   1.0
   ·   ─┬─
   ·    ╰─── Unexpected token 1.0
───╯
```

This particular example is not really meaningful. But, for example, it's a valid use case to call a function for some side effect, not for the return value.

```
fn foo(x:float) {
  print(x)
  x * 2.0
}
```

The reason of the failure is, if I understand correctly, the current parser allows an expression only either

1. in a statement (`let` statement, `fn` declaration, and `macro` declaration)
2. on the next line of a specific type of statement (`Expr::Let`, `Expr::LetTuple`, and `Expr::LetRec`)
3. as a single line in a `{ }` block

There can be several possible solutions, e.g. let a `Expr::Block` hold a vector of expressions.

In this pull request, I chose to tweak the rule 2; introduce "expression statement" and create a chain of expressions. 

Expression statement is typically defined as following in the case of a language that requires `;`. On the other hand, Mimium is semicolon-less, so, an expression and an expression statement look the same (except when a ';' is explicitly appended). However, I believe these two needs to be somehow distinguished in the implementation if the language has concept of expression and statement.

```
Expression ';'
```

This pull request's approach is incomplete, but seems to work at least to some extent. 

A disclaimer is that, this is based on the mimium's syntax I imagined wildly ([gist](https://gist.github.com/yutannihilation/eb3f3db5cd4a2f388bffd5bf856985c3)), so I might have some misunderstandings on the syntax and semantics!

### DIFF
--- a/mimium-lang/src/compiler/parser.rs
+++ b/mimium-lang/src/compiler/parser.rs
@@ -2,7 +2,9 @@ use crate::ast::*;
 use crate::types::{PType, Type, TypedId};
 use crate::utils::error::ReportableError;
 use crate::utils::metadata::*;
+use chumsky::combinator::Or;
 use chumsky::prelude::*;
+use chumsky::primitive::Just;
 use chumsky::Parser;
 mod token;
 use token::{Comment, Op, Token};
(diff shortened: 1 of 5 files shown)


mimium-lang/src/compiler/parser.rs
```diff
@@ -46,351 +48,412 @@ fn type_parser() -> impl Parser<Token, Type, Error = Simple<Token>> + Clone {
...
+    let ops: Vec<_> = ops.into_iter().map(optoken).collect();
+
+    lhs.then(choice(ops).then(rhs).repeated())
+        .foldl(move |x, ((op, opspan), y)| op_cls(x, y, op, opspan))
+        .boxed()
+}
+
+fn non_block_expr_parser<'a>(
+    expr: impl Parser<Token, WithMeta<Expr>, Error = Simple<Token>> + Clone + 'a,
+) -> impl Parser<Token, WithMeta<Expr>, Error = Simple<Token>> + Clone + 'a {
```
yutannihilation, line 133: By taking a argument of `impl Parser`, this function can take either 

1. a recursive parser passed from `recursive()`
2. a non-recursive parser like `literal_expr_parser()`

Because of 1., we can separate functions instead of putting all the code into the lambda of `recursive()`. By doing this, probably parser test can be written more easily.

yutannihilation (commented):
Some findings about quality of life with chumsky.

tomoyanonymous:
Splitting recursive parsers through `declare()` looks nice! That was what I was struggling with chumsky.

`Then` expr is what I reserved for the functionality of `ExprStmt` you made.

In the previous implementation, there were 3 cases assumed for the multi-line statement.

1. variable declaration w/ `let` (though previous spec did not require explicit `let`)
2. destructive assignment
3. function call which might have some side effects(this includes task scheduling expression with `@` operator)

I rejected other cases at the parser level, by splitting expr with fcall and non-fcall expression.
 
https://github.com/mimium-org/mimium/blob/dev/src/compiler/mimium.yy

However, considering pipe operator `|>` with side effects into account, maybe we should allow any sequence of expression at the parser level, and emit error or warning at the type checker(the first expression should have unit type).

Anyway, I was planning to revive destructive assignment and `Then` expr in the scheduler implementation after I completed tuple unpack impl.

Your draft and the previous implementation allows function definition with  `fn` keyword  inside block, because `let` statement and function definition statement are in the same level but I want to avoid it. This relates to a future implementation of module visibility between multiple files. We need to have some distinction between `Statement`(let, assign, and the others) and
 ‘TopLevelStatement‘ (statement and function definitions, and maybe module include in the future), for example.

Also, though currently not implemented yet, line break right after binary operators should be allowed for the long expressions, especially with the pipe operator. For this, we might modify the lexer-level syntax definition.

yutannihilation:
> `Then` expr is what I reserved for the functionality of `ExprStmt` you made.

Aw, I was wondering about this after I found `Then`. Got it.

> Anyway, I was planning to revive destructive assignment and `Then` expr in the scheduler implementation after I completed tuple unpack impl.

I see. Thanks for sharing the plan. Then, I'm closing this and will prepare for the upcoming changes. This was a good practice for me, but sorry for taking your time to explain the details. I hope I can make actual contribution around this in future!

> Your draft and the previous implementation allows function definition with `fn` keyword inside block, because `let` statement and function definition statement are in the same level but I want to avoid it

Yeah, I was aware of this gap between my implementation and the intended syntax and was thinking how to deal with it. It might be a good strategy to allow the chumsky parser to parse more than the actual syntax, and then validate the AST. I this way, probably mimium can show nicer error messages like "You cannot define function here". Also, no distinction might simplify the combination of chumsky parsers. However, it might make the validation step more complex.

tomoyanonymous:
> I was thinking how to deal with it. It might be a good strategy to allow the chumsky parser to parse more than the actual syntax, and then validate the AST. I this way, probably mimium can show nicer error messages like "You cannot define function here". 

That's fair. Maybe we can make custom parser error messages with `validate`  and  `try_map` in chumsky.

PR #25, opened 2024-08-19
Use StringInterner

yutannihilation:
This pull request addresses a part of #12. Probably this implementation is not what you expected, so I'd like to hear your thoughts.

I implemented this differently than otopoiesis in that this uses a thread local storage rather than extending `Context`. This idea comes from the Rust compiler. Let me explain a bit about it rather than about my implementation in this pull request.

It uses a `SessionGlobals` ([definition](https://github.com/rust-lang/rust/blob/804be74e3c4ec3646b448dfdb7a0fc31237a5818/compiler/rustc_span/src/lib.rs#L98-L110)) to intern symbols and spans.

``` rust
scoped_tls::scoped_thread_local!(static SESSION_GLOBALS: SessionGlobals);
```
([code](https://github.com/rust-lang/rust/blob/804be74e3c4ec3646b448dfdb7a0fc31237a5818/compiler/rustc_span/src/lib.rs#L175))

``` rust
pub fn with_session_globals<R, F>(f: F) -> R
where
    F: FnOnce(&SessionGlobals) -> R,
{
    SESSION_GLOBALS.with(f)
}
```
([code](https://github.com/rust-lang/rust/blob/804be74e3c4ec3646b448dfdb7a0fc31237a5818/compiler/rustc_span/src/lib.rs#L160-L165))

With this, the code is much simpler because these functions don't need to take a `Context` as an argument so that we don't need to pass a `Context` around. In the Rust compiler, conversion between `Symbol` and `&str` is implemented as following.

``` rust
impl Symbol {

    ...snip...

    pub fn intern(string: &str) -> Self {
        with_session_globals(|session_globals| session_globals.symbol_interner.intern(string))
    }

    ...snip...

    pub fn as_str(&self) -> &str {
        with_session_globals(|session_globals| unsafe {
            std::mem::transmute::<&str, &str>(session_globals.symbol_interner.get(*self))
        })
    }
```
([code](https://github.com/rust-lang/rust/blob/804be74e3c4ec3646b448dfdb7a0fc31237a5818/compiler/rustc_span/src/symbol.rs#L2311-L2328))

Note that, while this is thread-local, the source code says compilation can be done in parallel by making `SessionGlobals` a `Send`. (This pull request isn't implemented as such, though)

https://github.com/rust-lang/rust/blob/d0293c6cf22cb37f04ac90a52d1fc259362acc5b/compiler/rustc_interface/src/util.rs#L198-L201

I think an interner for span can be implemented in the same way because the Rust compiler does so. Regarding the interner for `Expr`, I have no idea at the moment.

Btw, the result of `cargo bench` is almost the same as the current HEAD of `dev`.


tomoyanonymous:
Setting the interner as thread_local-global variable looks reasonable.

> Btw, the result of cargo bench is almost the same as the current HEAD of dev

The current benchmark code is for measuring run-time performances. So it ignores a time consumed during the compilation. We need to add another benchmark to measure compilation time.

yutannihilation:
Thanks for the prompt response! I see. Then, I'll wait for #22 and finish this.

> The current benchmark code is for measuring run-time performances.

Ah, I missed it, sorry. I'll try to add some simple cases.

yutannihilation:
Result.

**Before**
```
test tests::parse::bench_many_symbols10 ... bench:  21,927,470.00 ns/iter (+/- 1,608,471.00)
test tests::parse::bench_many_symbols3  ... bench:   2,213,405.00 ns/iter (+/- 112,039.50)
test tests::parse::bench_many_symbols5  ... bench:   5,386,600.00 ns/iter (+/- 310,793.50)
```

**After**
```
test tests::parse::bench_many_symbols10 ... bench:  19,623,480.00 ns/iter (+/- 1,612,232.00)
test tests::parse::bench_many_symbols3  ... bench:   2,058,487.50 ns/iter (+/- 86,609.00)
test tests::parse::bench_many_symbols5  ... bench:   4,964,400.00 ns/iter (+/- 158,692.00)
```

yutannihilation:
The Rust compiler's implementation is smart in that it "prefill"s the symbols for builtin functions by generating such a `impl` for the interner by a proc-macro.

* https://github.com/rust-lang/rust/blob/804be74e3c4ec3646b448dfdb7a0fc31237a5818/compiler/rustc_span/src/symbol.rs#L110
* https://github.com/rust-lang/rust/blob/e3f909b2bbd0b10db6f164d466db237c582d3045/compiler/rustc_macros/src/symbols.rs#L317-L323

How is this possible? This is because `Symbol::new()` is `const fn`, which takes a `usize` of the actual index of the symbol in the storage.

https://github.com/rust-lang/rust/blob/804be74e3c4ec3646b448dfdb7a0fc31237a5818/compiler/rustc_span/src/symbol.rs#L2302-L2304

This way also works with string-interner.

``` rust
use string_interner::{backend::StringBackend, StringInterner};

fn main() {
    let earth_sym = 0;
    let air_sym = 3;
    let interner =
        <StringInterner<StringBackend<usize>>>::from_iter(["Earth", "Water", "Fire", "Air"]);

    println!("{}: {:?}", earth_sym, interner.resolve(earth_sym));
    println!("{}: {:?}", air_sym, interner.resolve(air_sym));
}
```
```
0: Some("Earth")
3: Some("Air")
```

But, I think I'll create a separate pull request. Adding a proc-macro might take some time.

yutannihilation:
Rebased to the tuple unpacking commit. Should be ready for review now.

PR #27, opened 2024-08-21
Use id-arena to store `Expr` and `Type`

yutannihilation:
Address  #12

## Overview

This pull request introduces the following storage into `SessionGlobals`

1. `expr_storage` to store `Expr`
2. `type_storage` to store `Type`
3. `span_storage` to store `Span` with the key of the ID of `Expr` or `Type`

As a result,

* Since `Expr` and `Type` are immediately converted into a copyable Id, so `Box` is no longer needed for recursive definitions
* Since all `Span`s are recorded in `span_storage`, `WithMeta` is no longer needed

### ID for `Span` of `TypePattern` and `TypedId`

Since `TypePattern` and `TypedId` has only one `Type`, I decided to use this ID for the key of the corresponding `Span`. However, of course, this is not accurate. For example, a `TypedId` represents a `Symbol` and a colon in addition to the `Type`. I guess this isn't a big problem at the moment, but the error message can be confusing because of this.

``` rust
    let x:float = 1.0
//        ^^^^^ 1
//      ^^^^^^^ 2
```

Maybe I could add something like `type_pattern_storage` and `typed_id_storage`, and use the key. But, I'd like to keep this pull request simple because this is already a bit complex.


yutannihilation:
Currently, the test fails with the following error. This is weird because, if I investigate manually, there's no difference between the actual `Expr`s of these two `ExprId`s.

```
---- compiler::parser::test::test_tuple stdout ----
thread 'compiler::parser::test::test_tuple' panicked at mimium-lang\src\compiler\parser\test.rs:266:5:
assertion `left == right` failed
  left: WithMeta(Tuple([ExprId(Id { idx: 3 }), ExprId(Id { idx: 4 })]), 0..10)
 right: WithMeta(Tuple([ExprId(Id { idx: 0 }), ExprId(Id { idx: 1 })]), 0..10)
```

tomoyanonymous:
Thanks. This really needed to address the type inference issue...
Having span as `BTreeMap` with `ExprId` is what I was expected for the implementation.

> Currently, the test fails with the following error. This is weird because, if I investigate manually, there's no difference between the actual Exprs of these two ExprIds.

`ExprId` should not derive `PartialEq` and `Eq` but should impl PartialEq manually by comparing actual `Expr` in the comparison.
There maybe the situation that we need to compare internal id but we can still do it by accessing internal `id_arena::Id<Expr>` directly.

tomoyanonymous:
`Pattern` and `Type` should be allocated on the session_global similarly...

yutannihilation:
> `ExprId` should not derive `PartialEq` and `Eq` but should impl PartialEq manually by comparing actual `Expr` in the comparison.

Oh, thanks for your insight!!! I'll confirm this tomorrow.

> `Pattern` and `Type` should be allocated on the session_global similarly...

I agree. I'm not immediately sure if these need dedicated storage for each, or a variant of `Expr` enum  (e.g. `Expr::TypePattern(TypePattern)`), but these probably need to be stored in session_globals somehow.

tomoyanonymous:
>I'm not immediately sure if these need dedicated storage for each, or a variant of Expr enum (e.g. Expr::TypePattern(TypePattern)

I more prefer to leaving `Pattern` as an individual recursive enum, Although we need to manage different storages, it ensures `Let` must hold `TypedPattern` in its left-hand, so it reduces the cost of validation.

yutannihilation:
If I implement `PartialEq` manually, it works fine. Thanks for the advice! (What I said "weird" was simply because I assumed wrongly that id-arena removes duplication automatically...)

tomoyanonymous:
I realized that the rust docs says
`The methods of this trait must be consistent with each other and with those of [PartialEq](https://doc.rust-lang.org/beta/std/cmp/trait.PartialEq.html).`

https://doc.rust-lang.org/beta/std/cmp/trait.PartialOrd.html

Now the key for `BtreeMap` is `NodeId`, not `ExprNodeId
` itself, so `ExprNodeId` should not derive `PartialOrd` and `Ord` too.

yutannihilation:
Thanks for catching!

yutannihilation:
I chose the direction to treat `Type` and `Expr` as the same. While this works, I'm getting to feel I was wrong. 

- `Type` can be used repeatedly. If one type and another have a same type, they should have the same ID rather than issuing a new one by `Arena::alloc()` every time.
- `Type` doesn't always have `Span`.

Probably I need to go back and forth several times.

tomoyanonymous:
> `Type` doesn't always have `Span`.

I think the BTreeMap from NodeId to Span can be extended to any kind of metadata. 

For example, I'm going to refactor the type inference code so that MIR generator later can refer to the map from `ExprNodeId` to `Type` after the entire inference finished.

Duplicating same values with different ids will not be an immidiate problem. 

Maybe, It is more clear that the Type values issued during the inference should be distinct from those of TypedPattern (Type inference context have an individual Arena) but it is ok to be allocated on the same Arena.

yutannihilation:
> Maybe, It is more clear that the Type values issued during the inference should be distinct from those of TypedPattern (Type inference context have an individual Arena) but it is ok to be allocated on the same Arena.

Thanks for sharing the plan. This sounds good to me! I agree this duplication isn't any problem (except for some possible performance overhead).

I think I don't understand the whole picture here, but probably I can finish this pull request today or tomorrow.

yutannihilation:
This should be basically ready for review now. I updated the main description.

One problem I'm seeing is that this pull request makes the benchmark tests fail with a segmentation fault. It seems this happens only on benchmark, so I don't think this is a blocker. If I run the same Mimium code by `cargo test` or by `cargo run`, nothing happens. I'll keep investigating.

```
warning: `mimium-lang` (lib) generated 40 warnings (run `cargo fix --lib -p mimium-lang` to apply 7 suggestions)
    Finished `bench` profile [optimized] target(s) in 0.04s
     Running benches/bench.rs (target/release/deps/bench-d16f0f1ff08fbac1)

running 1 test
error: bench failed, to rerun pass `-p mimium-lang --bench bench`

Caused by:
  process didn't exit successfully: `/home/yutani/repo/mimium-rs/target/release/deps/bench-d16f0f1ff08fbac1 'tests::parse::bench_many_symbols3' --exact --show-output --bench` (signal: 11, SIGSEGV: invalid memory reference)

 *  The terminal process "cargo 'bench', '--package', 'mimium-lang', '--bench', 'bench', '--', 'tests::parse::bench_many_symbols3', '--exact', '--show-output'" failed to launch (exit code: 101). 
 *  Terminal will be reused by tasks, press any key to close it. 
```

```
fn f0(x3:float) {
  let x3_0 = x3
  let x3_1 = x3
  let x3_2 = x3
  (((x3 * 0.1 + x3_0) * 0.1 + x3_1) * 0.1 + x3_2)
}



fn f1(x3:float) {
  let x3_0 = x3
  let x3_1 = x3
  let x3_2 = x3
  (((x3 * 0.1 + x3_0) * 0.1 + x3_1) * 0.1 + x3_2)
}



fn f2(x3:float) {
  let x3_0 = x3
  let x3_1 = x3
  let x3_2 = x3
  (((x3 * 0.1 + x3_0) * 0.1 + x3_1) * 0.1 + x3_2)
}


fn dsp() {
  f0(1.0) + f1(1.0) + f2(1.0)
}
```

tomoyanonymous:
> One problem I'm seeing is that this pull request makes the benchmark tests fail with a segmentation fault. It seems this happens only on benchmark, so I don't think this is a blocker. If I run the same Mimium code by cargo test or by cargo run, nothing happens. I'll keep investigating.

Did you tried with `--profile bench` option?

yutannihilation:
I didn't. Now I tried it, and it runs fine without any errors.

However, let me correct a bit. This code can reproduce the segfault.

- If I reduce the repetition of `emit_mir()`, it doesn't fail
- If I increase the amount of `content`, it fails with fewer repetition.

So, this is probably a real issue I need to address before merging. Probably, this might take some time. Please feel free to merge #28 first. I can retry after that...

``` rust
        #[test]
        fn t() {
            let content = "

...snip... (the code above)

";
            emit_mir(content).expect("ok");
            emit_mir(content).expect("ok");
            emit_mir(content).expect("ok");
            emit_mir(content).expect("ok");
            emit_mir(content).expect("ok");
            emit_mir(content).expect("ok");
            emit_mir(content).expect("ok");
            emit_mir(content).expect("ok");
            emit_mir(content).expect("ok");
        }
```

If I execute the above code with VS Code's Debug run, it points this line of `mimium-lang/src/compiler/mirgen/recursecheck.rs`. I have no idea in what situation this can happen.

image

tomoyanonymous:
Hmm, that test won't fail on macOS how many times I duplicate`emit_mir`. My fix in the review above does not change the behaviour too.

I guess it happens on `std::transmute<&Expr,&Expr>` in `to_expr` for extending lifetime? 

This transmute is done on [`rustc_span`](https://github.com/rust-lang/rust/blob/b5723af3457b9cd3795eeb97e9af2d34964854f2/compiler/rustc_span/src/symbol.rs#L2324) too but its content type is `&str`, not a recursive type. 
It causes some difference maybe?

tomoyanonymous:
This `get_expr` generates temporary `Option` struct and returning underlying reference. Extending lifetime of this temporary object will not be safe...

```rust
    pub fn get_expr(&self, expr_id: ExprNodeId) -> &Expr {
        self.expr_storage
            .get(expr_id.0)
            .expect("Unknown ExprNodeId")
    }
```

However, `id_arena` does not have `get_unchecked` method. 
`SlotMap` that I used in #28 have unchecked method, so should we switch to slotmap?

yutannihilation:
> Hmm, that test won't fail on macOS how many times I duplicate`emit_mir`.

Oh, good to know. I'm on a not-so-powerful Linux laptop, so maybe it matters.

> I guess it happens on `std::transmute<&Expr,&Expr>` in `to_expr` for extending lifetime?

I think so, too. I'll check if switching to `SlotMap` works. Thanks for the advice!

yutannihilation:
Quick update: I tried the same code on my Windows desktop and it works. However, if I increase the repetition to 50, it fails with the segfault.

yutannihilation:
Using `SlotMap` didn't solve the issue, unfortunately. Probably I should give up the unsafe implementation until I find the root cause.

The segfault is reproducible on CI.

https://github.com/tomoyanonymous/mimium-rs/actions/runs/10521271003/job/29151628740?pr=27#step:6:521

yutannihilation:
Okay, I give up avoiding the cloning costs. Now the benchmark doesn't fail.

Benchmark results:

**Before**
```
test tests::parse::bench_many_symbols10 ... bench:  22,420,550.00 ns/iter (+/- 2,571,804.00)
test tests::parse::bench_many_symbols3  ... bench:   2,301,567.50 ns/iter (+/- 56,266.75)
test tests::parse::bench_many_symbols5  ... bench:   5,573,355.00 ns/iter (+/- 183,844.00)
test tests::runtime::bench_multiosc10   ... bench:       3,429.20 ns/iter (+/- 109.64)
test tests::runtime::bench_multiosc15   ... bench:       5,100.33 ns/iter (+/- 125.19)
test tests::runtime::bench_multiosc5    ... bench:       1,748.98 ns/iter (+/- 48.91)
test tests::runtime::bench_multiosc7    ... bench:       2,396.01 ns/iter (+/- 54.89)
```


**After**
```
test tests::parse::bench_many_symbols10 ... bench:  16,587,370.00 ns/iter (+/- 2,167,060.00)
test tests::parse::bench_many_symbols3  ... bench:   1,954,627.50 ns/iter (+/- 68,536.00)
test tests::parse::bench_many_symbols5  ... bench:   4,715,255.00 ns/iter (+/- 472,429.50)
test tests::runtime::bench_multiosc10   ... bench:       3,365.50 ns/iter (+/- 109.28)
test tests::runtime::bench_multiosc15   ... bench:       4,992.07 ns/iter (+/- 113.04)
test tests::runtime::bench_multiosc5    ... bench:       1,681.22 ns/iter (+/- 44.98)
test tests::runtime::bench_multiosc7    ... bench:       2,359.02 ns/iter (+/- 39.85)
```

tomoyanonymous:
Test runs without problems on macOS too!
One thing I want to correct is the `ExprNodeId` and `TypeNodeId`. They are implementing `PartialEq` by hand, so it should not derive `PartialOrd` and therefore should not be real keys to slotmap.

A key for slotmap can be generated with `new_key_type!` macro, and by wrapping it with `ExprNodeId`, manual trait implementation can be omitted.

https://github.com/tomoyanonymous/mimium-rs/commit/b5e94bbdd704e3987b0214254417e4d7b2bb509f

yutannihilation:
Wow, thanks for the implementation! It seems I was a bit confused. I first tried `new_key_type!` directly on `ExprNodeId` and `TypeNodeId`, which doesn't work, and went a wrong direction. Your commit looks good to me.

tomoyanonymous:
Okay, then I merge it now. Really thanks for the great contribution!

yutannihilation:
Thanks for all your help!

PR #28, opened 2024-08-22
Implmented auto de-allocation of Closure

tomoyanonymous:
In the previous implementation, any closures allocated on the VM are "closed" on the return instruction, that upvalues hold by closures are replaced from the location on the stack to reference-counted heap(RC<RefCell<V>>).
This PR addresses the issue.

- Storage of the closures are now implemented with `slotmap` crate. This is mostly same as `generational-arena`, the value can be removed and managed without fragmentation of memory spaces(generatinal-arena crate is no longer maintained).
- Currently, some tests fails because the type for inline closures are not inferred correctly. Maybe I merge this after the type inference code are fixed.


tomoyanonymous:
quick note: this is needed for the scheduler implementation. Each time tasks are scheduled, anonymous closures are made and
scheduler holds its id. When finished execution,
the scheduler orders to release the closure to the VM.

PR #30, opened 2024-08-23
Refactor Type Inference based on a slotmap expression

### LINKED ISSUE #8: Implement Tuple(Array?) Type

tomoyanonymous:
close #8  #12

Refactored type inference code so that all the inference done at first and MIR generator later refers to the BTreeMap from Id of Expr, to Id of Type.

I added `TypeScheme(u64,u64)` in the `Type` but this is for the future implementation of the generic type inference.

- [ ] Fix bugs in the inference for the tuple pattern binding
- [ ] Fix false error in the state_tuple.mmm


mimium-lang/src/compiler/mirgen.rs
```diff
@@ -422,32 +405,16 @@ impl Context {
                         inst_refs.push(self.get_current_basicblock().0.len());
                         self.push_inst(Instruction::GetElement {
                             value: dst.clone(),
-                            ty: Type::Unknown.into_id(), // lazyly set after loops
+                            ty, // lazyly set after loops
```
yutannihilation, line 408: This `ty` is overwritten by `let (v, ty) = self.eval_expr(*e)?;`. It seems all the tests pass (except for `state_tuple`) by changing the variable name to avoid conflict.

``` patch
index 1c1dac6..160d6af 100644
--- a/mimium-lang/src/compiler/mirgen.rs
+++ b/mimium-lang/src/compiler/mirgen.rs
@@ -398,7 +398,7 @@ impl Context {
                 let dst = self.gen_new_register();
                 let mut inst_refs: Vec<usize> = vec![];
                 for (i, e) in items.iter().enumerate() {
-                    let (v, ty) = self.eval_expr(*e)?;
+                    let (v, elem_ty) = self.eval_expr(*e)?;
                     let ptr = if i == 0 {
                         dst.clone()
                     } else {
@@ -410,7 +410,7 @@ impl Context {
                             tuple_offset: i as u64,
                         })
                     };
-                    self.push_inst(Instruction::Store(ptr, v, ty));
+                    self.push_inst(Instruction::Store(ptr, v, elem_ty));
                 }
                 self.get_current_basicblock()
                     .0
```

tomoyanonymous:
The newly added example, highpass filter running correctly but lowpass filter runs wrong when cutoff frequency is less than about 5500Hz...

```rust
let pi = 3.14159535
let sr = 44100.0
fn phasor(freq){
  (self + freq/sr)%1.0
}
fn osc(freq){
  (phasor(freq)*2.0)-1.0
}

fn gen_hp_coeff(fc,q,fs){
    let omega = 2.0 * pi* fc / fs
    let alpha = sin(omega)/(q)
    let norm = 1.0 + alpha
    let a1 = -2.0 * cos(omega) / norm
    let a2 = (1.0 - alpha)/ norm
    let btmp = 1.0 + cos(omega)
    let b0 = btmp / (2.0*norm)
    let b1 = -btmp / norm
    let b2 = b0
    (a1,a2,b0,b1,b2)
}
fn gen_lp_coeff(fc,q,fs){
    let omega = 2.0 * pi* fc / fs
    let alpha = sin(omega)/q
    let norm = 1.0 + alpha
    let a1 = -2.0 * cos(omega) / norm
    let a2 = (1.0 - alpha)/ norm
    let btmp = 1.0 - cos(omega)
    let b0 = btmp / (2.0*norm)
    let b1 = btmp  / norm
    let b2 = b0
    (a1,a2,b0,b1,b2)
}
fn inner (x,a1,a2){
    let (ws, wss, _wsss) = self
    let w = x - a1*ws - a2*wss
    (w, ws, wss)
}
fn biquad(x,coeffs){
    let (a1,a2,b0,b1,b2) = coeffs;
	let (w,ws,wss) = inner(x,a1,a2);
	b0*w + b1*ws + b2*wss
}

fn dsp(){
    biquad(osc(440.0),gen_hp_coeff(2000.0,2.0,sr))
}
```

tomoyanonymous:
It seems that still type inference code failing intermediate type substitution, that is causing the bug.

By adding explicit type annotations for `mem_by_hand` in the code below, this code passes the test.

```rust
fn counter(){
    1.0+self
}
fn mem_by_hand(x){
    let (y,ys) = self
    (x,y)
}
fn dsp(){
    mem_by_hand(counter())
}
```

tomoyanonymous:
In this code, substitution for i(3)->i(6) is made first, and i(3)->float is made later.
Current substitution follows the intermediate id by order, and looks up the actual substitution.
The unification should be implemented with single linked list like done in the previous implementation in C++...

tomoyanonymous:
memo: `ena` crate, which is used in rustc internally, would be nice

https://thunderseethe.dev/posts/unification/

tomoyanonymous:
I guess all the type inference code is now working correctly but still biquad code behaves wrong when using lowpass filter...

yutannihilation:
Curious.

Probably this isn't related to the test failure, but it seems the generated MIR of this code still contains an unresolved intermediate variable (`i(6)`). Is this an expected limitation at the moment?

```
fn counter(){
    1.0+self
}
fn mem_by_hand(x){
    let (y,ys) = self
    (x,y)
}
fn dsp(){
    mem_by_hand(counter())
}
```

```
fn mem_by_hand [x]
upindexes:[]upper:0 state_size: 1
  block 0
  ...
  reg(16) := getelement reg(11), (number,i(6())), 1[0]
  reg(17) := store reg(16), reg(15), i(6())
  reg(18) := retfeed reg(11) (number,i(6()))
fn dsp []
upindexes:[]upper:0 state_size: 2
  block 0
  ...
  reg(22) := call reg(21) [reg(20)] ->(number,i(6()))
             popstateidx  1
  reg(23) := ret reg(22) (number,i(6()))
```

tomoyanonymous:
yeah, there was a unification bug in pattern binding and it was fixed now.
Now `fb_mem` fails (x and y returns same value)
```rust
fn counter(){
    1.0+self
}
fn mem_by_hand(x){
    let (y,ys) = self
    (x,y)
}
fn dsp(){
    mem_by_hand(counter())
}
```

however, 2-sample delay with self succeeds...
```rust
fn counter(){
    1.0+self
}

fn mem_by_hand(x){
    let (y,ys,_yss) = self
    (x,y,ys)
}

fn dsp(){
    let (x,y, ys) =mem_by_hand(counter())
    (x,ys)
}
```

tomoyanonymous:
This is weird because the result of the `fb_mem` test sometimes passes and sometimes fails, even if I do not modify source code... This maybe caused by unsafe operation in vm.rs

tomoyanonymous:
No, bytecodegen generates sometimes different code... why is the behaviour not deterministic?

```diff
mem_by_hand
nparams:1 nret: 2
upindexes: []  state_size: 2  
constants:  []
instructions:
  getstate   1 2
  mov        5-6 1-2
  mov        3-4 5-6
  mov        7 0
  mov        5 7
--   mov        6 3
++  mov        7 3
++  mov        6 7
  getstate   7 2
  setstate   5 2
  ret        7 2
```

tomoyanonymous:
I found the bug in the virtual register allocation in codegen and now the inconsistency of the bytecodes are gone(idk why at all) though`fb_mem` test still fails.

tomoyanonymous:
It seems that the passing tuple as an argument is not working correctly. If making the coeff argument to individual variables, low-pass filter works.

tomoyanonymous:
Now biquad code is finally working, though only `fb_mem` test still fails somehow.

If I chenge bytecodegen for `Returnfeed` like this, the code works well,
```diff
            mir::Instruction::ReturnFeed(new, rty) => {
                //for returning always 0 at t=0
--                let old = self.vregister.add_newvalue(&dst);
                let bytecodes_dst = bytecodes_dst.unwrap_or_else(|| funcproto.bytecodes.as_mut());
                let size = Self::word_size_for_type(*rty);
--              bytecodes_dst.push(VmInstruction::GetState(old, size));
                let new = self.find(new);
                bytecodes_dst.push(VmInstruction::SetState(new, size));
                let nret = Self::word_size_for_type(*rty);
--                Some(VmInstruction::Return(old, nret))
++               Some(VmInstruction::Return(new, nret))
            }
```

though `counter` code returns 1 instead 0 at time=0. It does not make big effect for most of signal processing, though I am wondering which is the appropriate spec

```rust
//Which should we return 0 or 1 at time=0?
fn counter(){
 self+1
}
```

yutannihilation:
> only `fb_mem` test still fails somehow

I still see `biquad_hp` test is also failing. Do you have any unpushed changes in your local? Or, does the result differ depending on the OS or arch...?

https://github.com/tomoyanonymous/mimium-rs/actions/runs/10559878999/job/29252186124#step:6:640

(I'm still trying to understand what's happening in `fb_mem`...)

tomoyanonymous:
> I still see biquad_hp test is also failing. Do you have any unpushed changes in your local? Or, does the result differ depending on the OS or arch...?
Sorry, I was going to remove `biquad_hp` test because it is difficult to assume correct answer...

yutannihilation:
Ah, I see. Sounds reasonable!

tomoyanonymous:
It was likely caused by the logic for setting multiple values to stack was wrong. Now it is ready to merge finally.

PR #31, opened 2024-08-23
Use predefined symbols

yutannihilation:
As `StringInterner` assigns the index sequentially, we can use predefined symbols by ensuring the strings are interned by a specific order. This is what I commented on https://github.com/tomoyanonymous/mimium-rs/pull/25#issuecomment-2296470211.

I think this is not a priority. I still need to figure out several remaining things.

* Is this useful at all?
* How can I make compilation fail if there are duplicated entries?
* ~~Does `Type` need similar trick? (`SlotMap` has `insert_with_key()`, so we can probably prefill it. But `Key` is probably not `const`)~~ (edit: `insert_with_key()` was not what I expected. It just inserts a pair of key and value, so it's not possible to determine the key before inserting.)

---

### Usage

``` rust
mimium_symbols! {
    EntryPoint {
        dsp,
        global: "_mimium_global",
    }

    BuiltinFn {
        neg,
        sin,
        cos,
    }

    SpecialFn {
        delay,
    }
}
```

Above code is expanded into the following

``` rust
pub mod symbols {
    pub mod entry_point {
        use crate::interner::Symbol;
        pub const DSP: Symbol = Symbol(0usize);
        pub const GLOBAL: Symbol = Symbol(1usize);
    }
    pub mod builtin_fn {
        use crate::interner::Symbol;
        pub const NEG: Symbol = Symbol(2usize);
        pub const SIN: Symbol = Symbol(3usize);
        pub const COS: Symbol = Symbol(4usize);
    }
    pub mod special_fn {
        use crate::interner::Symbol;
        pub const DELAY: Symbol = Symbol(5usize);
    }
}
impl crate::interner::SessionGlobals {
    pub(crate) fn prefill(&mut self) {
        for s in ["dsp", "_mimium_global", "neg", "sin", "cos", "delay"] {
            self.symbol_interner.get_or_intern(s);
        }
    }
}
```


tomoyanonymous:
The macros looks nice but feels a bit overkill for a current complexity of our codes at the same time.

Ordering pre-defined symbols linearly, it is not meaningful by itself because the does not affect to accessing speed and we do not iterate over the interned symbols sequentially.

In my view, because rustc treats so-so many symbols including not only language's keyword but also special functions used in other tools like clippy and language-server(I guess so but not sure), so this is meaningful for them.

Maybe when we started developing a language server(this is also what I want to try with rust) these macro could become useful.

yutannihilation:
Yeah, I was thinking the main point of this is not about performance, but that we can gather the definition in one place so that we can avoid inconsistency. (I planned to generate `get_builtin_fns()` etc as well).

But, after experimenting a bit, I agree this is an overkill. Let's close this and revisit when this is needed.

Thanks for taking a look!

PR #34, opened 2024-08-26
Improve the calculation of state size

yutannihilation:
It seems this slice might contain uninitialized values. This pull request is mainly to see if the error is reproducible on CI. (Actually, I haven't come up with how to fix this...)

https://github.com/tomoyanonymous/mimium-rs/blob/72b5d230fd4bcbe22dfb1ab41f087c00dc6926b5/mimium-lang/src/runtime/vm.rs#L305

If I dump the `vs` at the point shown below, the fourth value is not zero.

```
      ret        2 1
    mem_by_hand
    nparams:1 nret: 4
    upindexes: []  state_size: 4
    constants:  []
    instructions:
-->   getstate   1 4
      mov        9-12 1-4
      mov        5-8 9-12
      mov        13 0
      mov        9 13
      mov        13 5
      mov        10 13
      mov        13 6
      mov        11 13
      mov        13 7
      mov        12 13
      getstate   13 4
      setstate   9 4
      ret        13 4
    dsp
    nparams:0 nret: 2
```

```
0000000000000000000000000000000000000000000000000000000000000000
0000000000000000000000000000000000000000000000000000000000000000
0000000000000000000000000000000000000000000000000000000000000000
0000000100010001000000000000000100000000000000000000000000000001
```

### DIFF
--- a/mimium-lang/src/compiler/bytecodegen.rs
+++ b/mimium-lang/src/compiler/bytecodegen.rs
@@ -461,11 +461,11 @@ impl ByteCodeGenerator {
             }
             mir::Instruction::PushStateOffset(v) => {
                 let state_size = Self::calc_state_size(v) as i16;
-                Some(VmInstruction::SetStatePos(state_size))
+                Some(VmInstruction::ShiftStatePos(state_size))
             }
             mir::Instruction::PopStateOffset(v) => {
                 let state_size = Self::calc_state_size(v) as i16;
-                Some(VmInstruction::SetStatePos(0))
+                Some(VmInstruction::ShiftStatePos(-state_size))
             }
             mir::Instruction::GetState(ty) => {
                 let size = Self::word_size_for_type(*ty);
--- a/mimium-lang/src/compiler/bytecodegen.rs
+++ b/mimium-lang/src/compiler/bytecodegen.rs
@@ -162,9 +162,7 @@ impl ByteCodeGenerator {
         state_sizes
             .as_ref()
             .iter()
-            .map(|x| match x {
-                StateSize::Typed(size, ty) => *size * Self::word_size_for_type(*ty) as u64,
-            })
+            .map(|x| x.size * Self::word_size_for_type(x.ty) as u64)
             .sum()
     }
     fn get_binop(&mut self, v1: &Arc<mir::Value>, v2: &Arc<mir::Value>) -> (Reg, Reg) {
--- a/mimium-lang/src/compiler/bytecodegen.rs
+++ b/mimium-lang/src/compiler/bytecodegen.rs
@@ -158,7 +158,7 @@ impl ByteCodeGenerator {
             Type::Unknown => todo!(),
         }
     }
-    fn word_size_for_state_size<T: AsRef<[StateSize]>>(state_sizes: T) -> u64 {
+    fn calc_state_size<T: AsRef<[StateSize]>>(state_sizes: T) -> u64 {
         state_sizes
             .as_ref()
             .iter()
@@ -459,11 +459,11 @@ impl ByteCodeGenerator {
                     Self::word_size_for_type(*ty),
                 ))
             }
-            mir::Instruction::PushStateOffset(v) => Some(VmInstruction::ShiftStatePos(
-                Self::word_size_for_state_size(v) as i16,
-            )),
+            mir::Instruction::PushStateOffset(v) => {
+                Some(VmInstruction::ShiftStatePos(Self::calc_state_size(v) as i16))
+            }
             mir::Instruction::PopStateOffset(v) => Some(VmInstruction::ShiftStatePos(
-                -(Self::word_size_for_state_size(v) as i16),
+                -(Self::calc_state_size(v) as i16),
             )),
             mir::Instruction::GetState(ty) => {
                 let size = Self::word_size_for_type(*ty);
@@ -604,7 +604,7 @@ impl ByteCodeGenerator {
         fidx: usize,
     ) -> (Symbol, vm::FuncProto) {
         log::trace!("generating function {}", mirfunc.label.0);
-        let state_size = Self::word_size_for_state_size(mirfunc.get_state_sizes());
+        let state_size = Self::calc_state_size(mirfunc.get_state_sizes());
         let mut func = vm::FuncProto {
             nparam: mirfunc.args.len(),
             nret: Self::word_size_for_type(
--- a/mimium-lang/src/compiler/bytecodegen.rs
+++ b/mimium-lang/src/compiler/bytecodegen.rs
@@ -164,7 +164,6 @@ impl ByteCodeGenerator {
             .iter()
             .map(|x| match x {
                 StateSize::Typed(size, ty) => *size * Self::word_size_for_type(*ty) as u64,
-                StateSize::Word(size) => *size,
             })
             .sum()
     }
--- a/mimium-lang/src/compiler/bytecodegen.rs
+++ b/mimium-lang/src/compiler/bytecodegen.rs
@@ -158,7 +158,7 @@ impl ByteCodeGenerator {
             Type::Unknown => todo!(),
         }
     }
-    fn calc_state_size<T: AsRef<[StateSize]>>(state_sizes: T) -> u64 {
+    pub fn calc_state_size<T: AsRef<[StateSize]>>(state_sizes: T) -> u64 {
         state_sizes
             .as_ref()
             .iter()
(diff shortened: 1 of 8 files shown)


yutannihilation:
Hmm, it seems this doesn't happen on CI. But, it seems this occurs reliably on my Windows machine. This might be platform dependent.

yutannihilation:
It seems 10 times was not enough.

yutannihilation:
It seems the calculation of `state_size` is not accurate. `mir::Function` records the number of states, not the word size, and multiply it by the word size of the return type.

**MIR**

https://github.com/tomoyanonymous/mimium-rs/blob/72b5d230fd4bcbe22dfb1ab41f087c00dc6926b5/mimium-lang/src/compiler/mirgen.rs#L491-L493

https://github.com/tomoyanonymous/mimium-rs/blob/72b5d230fd4bcbe22dfb1ab41f087c00dc6926b5/mimium-lang/src/compiler/mirgen.rs#L341-L349

**bytecode**

https://github.com/tomoyanonymous/mimium-rs/blob/72b5d230fd4bcbe22dfb1ab41f087c00dc6926b5/mimium-lang/src/compiler/bytecodegen.rs#L596-L597

In this case,

```
fn counter(){
    1.0+self
}

fn mem_by_hand(x){
    let (y,ys,yss,_ysss) = self
    (x,y,ys,yss)
}

fn dsp(){
    let (x,y,ys,yss) = mem_by_hand(counter())
    (x,yss)
}
```

At the stage of MIR, `count` and `mem_by_hand`'s `state_size`s are both 1. The size of `dsp` is `2`, which is the sum of the sizes of these two functions.

```
fn counter []
upindexes:[]upper:0 state_size: 1
...

fn mem_by_hand [x:number]
upindexes:[]upper:0 state_size: 1
...

fn dsp []
upindexes:[]upper:0 state_size: 2
...
```

At the stage of bytecode, the `state_size` of `dsp` is calculated as `(MIR state_size) * nret = 2 * 2 = 4`. However, I think this should be `1 + 4 = 5` (am I correct...?).

```
counter
nparams:0 nret: 1
upindexes: []  state_size: 1
  ...

mem_by_hand
nparams:1 nret: 4
upindexes: []  state_size: 4
  ...

dsp
nparams:0 nret: 2
upindexes: []  state_size: 4
  ...

```

tomoyanonymous:
Yeah, size for dsp should be 5. State size at MIR should ideally be the collection of typeid...

yutannihilation:
Thanks, then I'll try modifying the representation of state size at MIR.

mimium-lang/src/compiler/bytecodegen.rs
```diff
@@ -452,8 +459,14 @@ impl ByteCodeGenerator {
...
-            mir::Instruction::PushStateOffset(v) => Some(VmInstruction::ShiftStatePos(*v as i16)),
-            mir::Instruction::PopStateOffset(v) => Some(VmInstruction::ShiftStatePos(-(*v as i16))),
+            mir::Instruction::PushStateOffset(v) => {
+                let state_size = Self::calc_state_size(v) as i16;
+                Some(VmInstruction::SetStatePos(state_size))
+            }
+            mir::Instruction::PopStateOffset(v) => {
+                let state_size = Self::calc_state_size(v) as i16;
+                Some(VmInstruction::SetStatePos(0))
+            }
```
yutannihilation, line 469: Now this pull request succeeded to get rid of errors by breaking the concept of push/pop. I'm not sure if this direction is right (especially because I don't understand how `make_delay()` works). However, I couldn't come up with a nice way to calculate the relative offset in `emit_fncall()`.

Maybe this is something that I should not mess up unless I figure out the higher plans... :thinking:

tomoyanonymous:
Yeah, the instructions related to stateoffset are the most unique and difficult part of the mimium, I'm gonna explain here briefly but, in short, having offset as relative value is the key part of a design in this VM.

This is the overall structure of the VM (the figures are from unpublished paper under review now, but it's single-blind so I think it's OK)
vmmodel
Basically, the compiler tries to collect the required memory space for any stateful function(which uses `self` or `delay`, and the functions which call statefull functions too) as possible as it can determine at compile time.
Such internal state values(ring buffer of just single value) are allocated on linear memory space.

For example, let us think the code for delay with feedback and some function uses it twice, and the dsp function call that function twice again, implemented like this.
```
fn fbdelay(x,fb,dtime){
x + delay(1000,self,dtime)*fb
}
fn twodelay(x,dtime){
    fbdelay(x,dtime,0.7)
      +fbdelay(x,dtime*2,0.8)
}
fn dsp(x){
    twodelay(x,400)+twodelay(x,800)
}
```
The bytecode for this code should be like this. (Sorry for the format are different from the real compiler)
```
CONSTANTS:[0.7,2,0.8,400,800,0,1]
fn fbdelay(x,fb,dtime) state_size:2
MOVE       3 0   //load x
GETSTATE   4     //load self
SHIFTSTATE 1     //shift Spos
DELAY      4 4 2 //delay(_,_,_)
MOVE       5 1   // load fb
MULF       4 4 5 //delayed val *fb
ADDF       3 3 4 // x+
SHIFTSTATE -1    //reset SPos
GETSTATE   4     //prepare result
SETSTATE   3     //store to self
RETURN     4 1   //return previous self

fn twodelay(x,dtime) state_size:4
MOVECONST  2 5 //load "fbdelay" prototype
MOVE       3 0
MOVE       4 1
MOVECONST  5 0 //load 0.7
CALL       2 3 1
SHIFTSTATE 2   //2=state_size of fbdelay
MOVECONST  3 5 //load "fbdelay" prototype
MOVE       4 0
MOVECONST  5 1 //load 2
MULF       4 4 5
MOVECONST  5 0 //load 0.7
CALL       3 3 1
ADDF       3 3 4
SHIFTSTATE -2
RETURN     3 1

fn dsp (x)
MOVECONST  1 6 //load "twodelay" prototype
MOVE       2 0
MOVECONST  3 3 //load 400
CALL       1 2 1 
SHIFTSTATE 4   //4=state_size of twodelay
MOVECONST  2 6
MOVE       2 3 //load "twodelay" prototype
MOVE       3 0
MOVECONST  3 4 //load 400
CALL       2 2 1
ADD        1 1 2
SHIFTSTATE -4
RETURN     1 1
```

By calling `twodelay`, position for state storage move along like this.
fbdelay_spos

The point is, by having offset in the instruction as a relative value,  `fbdelay` and `twodelay` do not need to care anything about where are they called from.

When we use closure i.e. when the functions are passed as value for higher-order function, the compiler can not determine state storage size at the compile time. So, when we use `callcls` operation, the VM switches statestorage from the global storage, to the other storage allocated on a closure.

We should add some tests

yutannihilation:
Thanks a lot! This really helps. I was actually wondering what happens on such a code with nested feedbacks.

> The point is, by having offset in the instruction as a relative value, `fbdelay` and `twodelay` do not need to care anything about where are they called from.

I got the point.

> So, when we use `callcls` operation, the VM switches statestorage from the global storage, to the other storage allocated on a closure.

This is also good to know.

I'm not sure if I can figure out how to make the implementation sound in all these aspects within this pull request, but I'll try.

yutannihilation:
Comparing that ideal bytecode and the ones generated by the current implementation, to me, it seems one difference is that the former emits `SHIFTSTATE` immediately after `CALL`, while the latter does it before the next call. I'm not sure if these are semantically different, but it works if I implement as such.

The bytecode is not very nice in that the last call of the block always emits a useless shift, though.

```
fn counter(){
  1.0+self
}

fn dsp(){
  counter() + counter() + counter()
}
```

```
dsp
nparams:0 nret: 1
upindexes: []  state_size: 3  
constants:  [1]
instructions:
  movc       0 0
  mov        1 0
  call       1 0 1
  shiftsttpos 1
  movc       2 0
  mov        3 2
  call       3 0 1
  shiftsttpos 1
  addf       0 1 3
  movc       1 0
  mov        2 1
  call       2 0 1
  shiftsttpos 1     <-- 
  addf       0 0 2
  shiftsttpos -3    <--
  ret        0 1
```

yutannihilation:
Now the implementation is more straightforward and should match with the current logic, although I still don't understand several points (e.g. calculation of `coffset` in `make_delay()`). I'd appreciate if you take a look when you have time, but no hurry at all!

tomoyanonymous:
Thanks a lot! It looks okay.

> The bytecode is not very nice in that the last call of the block always emits a useless shift, though.

This might be why I made `coffset` to combine redundunt shiftstate operations. 
Theoritically, Shiftstate operation can be swapped to adjoint insructions other than Getstate, Setstate, and Call. And if there's no these 3 operations between 2 Shiftstate operations, they can be combined.

I tried to solve this while emitting MIR but now I feel it is more natural to implement it as an optimization pass after complete MIR has emitted.

yutannihilation:
Thanks!

> This might be why I made `coffset` to combine redundunt shiftstate operations.

Ah, got it. It sounds a nice idea.

> I tried to solve this while emitting MIR but now I feel it is more natural to implement it as an optimization pass after complete MIR has emitted.

Agreed!

PR #36, opened 2024-08-30
Parse multiple white spaces at end

yutannihilation:
This pull request fixes a minor issue I found during playing with various codes.

The current lexer fails to parse the white spaces at end without line break. Among the following cases, only case 2 fails at the lexer stage.

1. `fn dsp() {\n  0.0\n} `: single white space at end
2. `fn dsp() {\n  0.0\n}  `: multiple white spaces at end
3. `fn dsp() {\n  0.0\n}  \n`: multiple white spaces at end with a linebreak

```
thread 'main' panicked at mimium-cli/src/main.rs:41:54:
called `Result::unwrap()` on an `Err` value: [ParseError(Simple { span: 19..19, reason: Unexpected, expected: {Some('^'), Some(';'), Some('\r'), Some('\n'), Some('('), Some('|'), Some('-'), Some('/'), Some('"'), Some(':'), Some('*'), Some('<'), Some('!'), Some('='), Some('0'), Some('.'), Some('+'), Some('}'), Some('['), Some(')'), Some('%'), Some(','), Some('&'), Some('{'), Some(']'), Some('>')}, found: None, label: None })]
```

It seems `repeated()` is needed in `padded_by()` to handle such white spaces. Unlike `padded()`, `padded_by()` doesn't apply `repeated()`.

(But, I couldn't understand how multiple white spaces are handled by the current parser without this `repeated()`... It's probably that these white spaces are somehow absorbed by one of the other subparsers, but I don't know which.)

### DIFF
--- a/mimium-lang/src/compiler/parser/lexer.rs
+++ b/mimium-lang/src/compiler/parser/lexer.rs
@@ -121,10 +121,12 @@ pub fn lexer() -> impl Parser<char, Vec<(Token, Span)>, Error = Simple<char>> {
         .or(linebreak)
         .recover_with(skip_then_retry_until([]));
 
+    let whitespaces = one_of(" \t\u{0020}").repeated();
+
     token
         .map_with_span(|tok, span| (tok, span))
         // .padded_by(comment_parser().repeated())
-        .padded_by(just(' ').or(just('\t').or(just('\u{0020}'))).or_not())
+        .padded_by(whitespaces)
         .repeated()
         .then_ignore(end())
 }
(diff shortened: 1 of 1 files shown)


tomoyanonymous:
> (But, I couldn't understand how multiple white spaces are handled by the current parser without this repeated()... It's probably that these white spaces are somehow absorbed by one of the other subparsers, but I don't know which.)

Maybe, each token can have single whitespace before and after themselves, then up to 2 whitespaces are allowed by the repeated tokens?
I think I mixed up the precedence of combinator, this is the correct implementation. Thanks!

yutannihilation:
> Maybe, each token can have single whitespace before and after themselves, then up to 2 whitespaces are allowed by the repeated tokens?

Ah, it makes sense. Thanks!

PR #37, opened 2024-08-30
Implement Destructive Assignment

tomoyanonymous:
This PR implements destructive assignment expression.

With this implementation, stateful object with closure can be expressed like below.

```rust
fn makecounter(){
    let x = 0.0
    let countup = | |{
        let res = x
        x = x+1.0
        res 
    }
    countup
}
let myc = makecounter();
fn dsp(){
    myc()
}
```

This PR also contains some parser refactoring previously attempted in #23.

### DIFF
--- a/mimium-lang/src/compiler/mirgen.rs
+++ b/mimium-lang/src/compiler/mirgen.rs
@@ -344,6 +344,43 @@ impl Context {
         };
         Ok(v)
     }
+    fn eval_assign(
+        &mut self,
+        name: Symbol,
+        src: VPtr,
+        t: TypeNodeId,
+        span: &Span,
+    ) -> Result<(), CompileError> {
+        match self.lookup(&name) {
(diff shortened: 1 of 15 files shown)


mimium-lang/src/compiler/mirgen.rs
```diff
@@ -347,6 +344,43 @@ impl Context {
...
+                Value::Argument(_i, _a) => Err(CompileError(
+                    CompileErrorKind::AssignmentToArg,
+                    span.clone(),
+                )),
+                _ => {
+                    self.push_inst(Instruction::Store(v.clone(), src, t));
+                    Ok(())
+                }
+            },
+            LookupRes::UpValue(_level, upv) => {
```
tomoyanonymous, line 365: I think I'm ready to merge it but it is still unclear whether deeply-nested closure and destructive assignments are treated correctly.
Level information is not used compared to the case of getupvalue.

PR #39, opened 2024-09-02
implement `now` (runtime-value)

tomoyanonymous:


tomoyanonymous:
I think the external closure mechanism is working, I'm gonna implement syntactic sugar `now` in another PR.

PR #41, opened 2024-09-03
Implement Scheduler Function

tomoyanonymous:
This PR adds a new built-in function ` _mimium_schedule_at` for a basic scheduler mechanism.
With this function, "temporal recursion" pattern can be expressed.
```rust
let x = 0.0

fn updater(){
    x = x + 1.0
    _mimium_schedule_at(now+1.0,updater)
}
let _ = _mimium_schedule_at(1.0,updater)
fn dsp(){
    x
}
```

This PR also fixes some minor issues.
- the upvalues referred by multiple closures are now correctly shared between closure instances.
  - this may cause some additional costs for the runtime performance.
- the bugs in recursion check are fixed to correctly compile the temporal recursion code.

In the succeeding PR, these tasks should be finished.

- auto deallocation of scheduled closures after the execution 
- add a syntax sugar for `_mimium_schedule_at` with `@` operator
- allow function call without return values in the parser with`Then` expr (currently we need to bind empty value with let)
- for the further abstraction, the language spec should allow that defining recursive function inside a function definition.(this is why `tests/mmm/scheduler_counter.mmm` fails now)
  - I'm gonna add `letrec` syntax for the parser, not allowing `fn(){}` syntax inside function definition
  - or, we can parse all let definition as letrec expr because we later convert non-recursive letrec to normal let anyway

```rust
fn makecounter(){
    let x = 0.0
    let updater = | |{
        x = x+1.0
    }
    let gen = | |{
        let _ = _mimium_schedule_at(now+1.0, updater)
        updater()
    }
    let getter = | | {x}
    let _ = _mimium_schedule_at(1.0,gen) // initialize a repetitive updater action
    getter
}
let x_getter = makecounter();
fn dsp(){
    x_getter()
}
```


mimium-lang/src/runtime/vm.rs
```diff
@@ -76,33 +71,57 @@ impl StateStorageStack {
...
 }
+type SharedUpValue = Rc<RefCell<UpValue>>;
 impl From<OpenUpValue> for UpValue {
     fn from(value: OpenUpValue) -> Self {
         Self::Open(value)
     }
 }
 
+#[derive(Default)]
+struct LocalUpValueMap(Vec<(Reg, SharedUpValue)>);
```
tomoyanonymous, line 89: This `LocalUpValueMap` is equivalent to the `OpenBroker` in the Lua Interpreter with Rust Tutorial.

https://github.com/WuBingzheng/build-lua-in-rust/blob/main/listing/ch09.closure/src/vm.rs#L85

They use binary search for shared-value search but I used iterator::find because the number of upvalues are usually small(need detailed comparison with the benchmark)

yutannihilation:
I'd like to understand the semantics of `@` a bit more precisely. This uses `*when <= now`, not `*when <= now && *when + 1 > now`, which means the scheduled function is executed even if the specified schedule already past.

https://github.com/tomoyanonymous/mimium-rs/blob/197e8db75e5123907b5ffb493cd6f25c4614498e/mimium-lang/src/runtime/scheduler.rs#L48-L58

For example, it seems these code give the same result. is this by design?

```
fn updater(){
    x = x + 1.0
    _mimium_schedule_at(now+1.0,updater)
}
```
```
// past
fn updater(){
    x = x + 1.0
    _mimium_schedule_at(now-1.0,updater)
}
```
```
// fixed
fn updater(){
    x = x + 1.0
    _mimium_schedule_at(0.0,updater)
}
```

tomoyanonymous:
We treat 3 time points here, let's say, 

- the time when `_mimium_schedule_at` is executed (`t_scheduled`)
- the time specified as an argument of `mimium_schedule_at` (`t_reserved`)
- the time when the task is actually executed (`t_executed`)

In the current synchronous scheduler implementation, the scheduled task at `(now +1.0)`  will be executed at 1 samples ahead of current time deterministically. (`t_scheduled < t_reserved = t_executed`)
However, in the future implementation, task scheduler may works on the other thread asynchronously and, in that implementation, scheduled task might be executed with a little bit delay  (`t_scheduled < t_reserved <= t_executed`).

Additionally, current `execute_task` implementation runs only one task even if multiple task scheduled at the same time is in the queue. This is a bug (thanks for pointing out). This is why the 3 codes above resulted in the same.

If we fixed this bug, `_mimium_schedule_at(t,task) ` when `t <= now` (`t_scheduled >= t_reserved`), the scheduler tries to repeat popping the tasks and causes infinite loop indeed, this case should emit a runtime error **when scheduled (not when executed)** . (Practically, the runtime should just ignore the task and continue the execution)

yutannihilation:
Ahh, thanks, makes sense!

> If we fixed this bug, `_mimium_schedule_at(t,task) ` when `t <= now` (`t_scheduled >= t_reserved`), the scheduler tries to repeat popping the tasks and causes infinite loop indeed,

Actually, I was wondering why it doesn't cause infinite loop... Now I get the point.

PR #42, opened 2024-09-03
Reserve index for dsp

yutannihilation:
The original motivation of this pull request was to add a validation to ensure the program has `dsp` function. After some experiment, I think the index of the `dsp` function can be predetermined just like `_mimium_global`. By fixing the index, the runtime doesn't need to track `dsp_i`.

I come up with the following two options, and I chose option 1 because this seemed relatively simple.

1. Add a flag to `Function` to indicate if the function is defined
2. Add a dedicated field (e.g. `dsp_function: Option<Function>`) to `Mir`

But, this makes the implementation slightly complex. `mirgen` needs to care a bit more about the status of the function. Does this look good to you?

Remaining questions:

* Is this implementation also sound with the cases of compiling libraries?
* Are there any case that a source code has multiple `dsp` functions?

### DIFF
--- a/mimium-audiodriver/src/backends/cpal.rs
+++ b/mimium-audiodriver/src/backends/cpal.rs
@@ -5,6 +5,7 @@ use crate::driver::{Driver, RuntimeData, SampleRate, Time};
 use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
 use cpal::{self, BufferSize, StreamConfig};
 use mimium_lang::interner::{Symbol, ToSymbol};
+use mimium_lang::mir::FN_INDEX_DSP;
 use mimium_lang::runtime::vm::{self, ExtClsType, Machine, ReturnCode};
 use ringbuf::traits::{Consumer, Producer, Split};
 use ringbuf::{HeapCons, HeapProd, HeapRb};
@@ -35,7 +36,6 @@ impl Default for NativeDriver {
 //Runtime data, which will be created and immidiately send to audio thread.
 struct NativeAudioData {
     pub vmdata: RuntimeData,
-    dsp_i: usize,
     dsp_ochannels: usize,
     buffer: HeapCons<f64>,
     localbuffer: Vec<f64>,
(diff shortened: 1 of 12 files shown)


tomoyanonymous:
We need to check whether the `dsp` function exists in the program when we really need to run it.
So, I think MIR do not need to distinct `dsp` and other functions unless we make some dsp-specific optimization in the future.

> Is this implementation also sound with the cases of compiling libraries?
Maybe yes. If we could obtain type signature and name of top-level functions in the other files, we can compile multiple files in parallel and link bytecodes, so not all MIR and Program have `dsp` function.

I put `dsp_i` in `RuntimeData`, not on `Program` with `Option` to avoid an additional condition check on runtime because it is invoked 44100 times/sec.

However, having `dsp_i` in `Program` and use it with `unwrap_unchecked` would be a balanced point of code's complexity and a run-time performance.

Another reason that I prefer to having `dsp_i` is, maybe I change the definition of the Program to "truly linked" program in the future: just a single array of bytecodes, not a array of functions, which have array of bytecodes like current implementation. Current structure is because it is based on Lua interpreter, whose functions are invoked by host program indeterminantly but in our use-cases, only a global context and `dsp` function are needed in most cases.

If we changed the definition of the program, `dsp_i` becomes just a index of the entire bytecodes, and it cannot be fixed location.

> Are there any case that a source code has multiple dsp functions?
I don't think so!

yutannihilation:
> Another reason that I prefer to having `dsp_i` is, maybe I change the definition of the Program to "truly linked" program in the future: just a single array of bytecodes, not a array of functions, which have array of bytecodes like current implementation.

Ah, makes sense! The program definitely need to hold the information of the location of `dsp` function. Thanks for clarifying. Then, I'm closing this pull request.

> We need to check whether the `dsp` function exists in the program when we really need to run it.

I'd argue this can be a compile error, not a runtime error. The compiler should know whether it's compiling an executable or a library. Just like `cargo` complains if we try to compile a binary crate without a `main` function (unless `#![no_main]` is specified). I'm not sure if this is too early to consider, though.

tomoyanonymous:
> I'd argue this can be a compile error, not a runtime error. The compiler should know whether it's compiling an executable or a library.
Ah, I got the motivation. Indeed it's in between compilation error and runtime error because it can be detected before launching VM but it's reason can't be associated to the specific location of the source code.
Currently the compiler boxing all the error to `Box<dyn ReportableError>` but there are errors without span like `NoDSPFnFound` or `FileNotFound`, in addition to the real runtime error like stack-overflow(though most of real runtime errors can be treated with "panic and exit" solution in most cases.)

BTW, Faust has "component" syntax, that is for using other file's `process` (equivalent to `dsp` of mimium) as a library function, though it is rarely used in real world.

https://faustdoc.grame.fr/manual/syntax/#component-expression

This may be useful in mimium too, and is like what you said "multiple dsp functions" situation.

yutannihilation:
> BTW, Faust has "component" syntax, that is for using other file's `process` (equivalent to `dsp` of mimium) as a library function, though it is rarely used in real world.

Sounds interesting!

Issue #10, opened 2024-08-04
Implement Scheduler


tomoyanonymous:
Because the closure and higher-order function can be realized in this implementation, Temporal recursion can be extended to more functional style, not just updating variable over time with destructive assignment by wrapping value like this.

```rust
fn generator(init:float, interval:float){
  let v = init // this v is captured and closed with recursive "updater" function
 let updater = | |{
     v = v+1
     updater()@(now+interval)
 }
 updater()@0;
 v
}
```

tomoyanonymous:
subtasks:

- implement closure auto de-allocation
- assign & then expression
- implment `@` operator as syntax sugar
- implement runtime value `now`
- implement logical task scheduler

tomoyanonymous:
Now the work is done!

PR #44, opened 2024-09-06
Use cargo-dist to distribute the release binaries

yutannihilation:
I guess you'll want to distribute the binary of `mimium-cli` when cutting a release so that those who don't know how to `cargo run` can also try it easily. I use [cargo-dist](https://opensource.axo.dev/cargo-dist/book/introduction.html) for such a purpose in [my project](https://github.com/yutannihilation/savvy) and it's quite comfortable so far.

This pull request adds the minimum settings for using cargo-dist. After adding this, as explained in the introduction page of cargo-dist, you can just push a tag to invoke the CI build. Then, the CI will publish a GitHub release with the binaries.

``` sh
git commit -am "release: 0.0.1"
git tag "v0.0.1"
git push v0.0.1
```

Here's an actual example I tried on my forked repo:

* release: https://github.com/yutannihilation/mimium-rs/releases/tag/v0.0.1-alpha.1
* action: https://github.com/yutannihilation/mimium-rs/actions/runs/10731270758/job/29761092415

While the diff in this pull request looks overwhelming, `.github/workflows/release.yml` is auto-generated by `cargo dist init`, so I think you can just ignore.

Let me explain the rest of the changes.

* `[workspace.metadata.dist]` and `[profile.dist]` are also added by `cargo dist init` and I don't modify anything here.
  * cargo-dist can also generate various installers like a shell script to `curl | sh`. Homebrew and MSI are useful, but probably it's too early for now, and I haven't used these...
* `[workspace.metadata.dist.dependencies.apt]` is necessary because cpal requires libasound.
  * It seems macOS runner and Windows runner doesn't need additional system dependencies. cpal's CI [installs ASIO SDK](https://github.com/RustAudio/cpal/blob/cc8ba755438ea1d186e024fec68db9390594e495/.github/workflows/cpal.yml#L160-L166), but I don't know if it's necessary or not.
* `[workspace.metadata.dist.github-custom-runners]` is probably not necessary. This is just I felt ubuntu-20.04, cargo-dist's default runner, is a bit outdated.
* `repository` setting is necessary. It seems this is used for the URLs of the artifacts in the GitHub Release. (So, I needed to use `https://github.com/yutannihilation/mimium-rs` on my forked repo)

Note that, I haven't used cargo-dist to the binaries that links to a library, so I'm not fully sure if this binary works without problems on any machines. At the moment, I confirmed the Linux binary works on my Manjaro Linux. It seems this requires users to install libasound, so some instructions might be needed. I can also test on Windows later. For macOS, I don't have a macOS machine, so I'm not sure (but I think it works without any additional setup).

``` sh
❯ ldd mimium-cli-x86_64-unknown-linux-gnu/mimium-cli
	linux-vdso.so.1 (0x0000715f0360b000)
	libasound.so.2 => /usr/lib/libasound.so.2 (0x0000715f0311c000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x0000715f035b5000)
	libm.so.6 => /usr/lib/libm.so.6 (0x0000715f0302d000)
	libc.so.6 => /usr/lib/libc.so.6 (0x0000715f02e3c000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x0000715f0360d000)
```


tomoyanonymous:
I'm going to merge this to try pre-release at the point of #10 is finished.
The built binary worked on my M1 Macbook air, so it gonna be okay.

yutannihilation:
> The built binary worked on my M1 Macbook air

Great! I forgot to report here, but it worked on my Windows as well. Thanks for merging.

PR #46, opened 2024-09-07
Implement syntactic sugar `@`

yutannihilation:
This pull request add a functionality to parse `@` as `_mimium_schedule_at`.

The things that are still unclear to me:

* This pull request simply treats `@` as an alias to `_mimium_schedule_at` just like other operators. As a syntactic sugar, does this require some special consideration?
* Does `@` precedes all other infix operators? (e.g. `foo@1^2` is translated as `(foo@1) ^ 2` (so this is an invalid expression)?)


tomoyanonymous:
Thanks! I was working to add this syntax sugar, with solving the problem that the single function execution without return value can not be parsed. However it's going so messy, I'll merge this first.

> Does @ precedes all other infix operators? (e.g. foo@1^2 is translated as (foo@1) ^ 2 (so this is an invalid expression)?)

`@` operator has the lowest precedence because it never returns values and then never used multiple times in a single expression.

yutannihilation:
> However it's going so messy, I'll merge this first.

Ah, I see. After creating this pull request, while treating it as an usual op is simple, I was wondering if `@` should be a part of function call here. I guess the parser needs larger modification to allow a function call without return value anyway, so I'm fine either if this gets merged or not. Please do which is easier to you.

https://github.com/tomoyanonymous/mimium-rs/blob/4912828a82c99e308ebc2c77b9fb8e9588161b1a/mimium-lang/src/compiler/parser.rs#L229-L245


> `@` operator has the lowest precedence because it never returns values and then never used multiple times in a single expression.

Thanks for clarifying!

yutannihilation:
> I was wondering if `@` should be a part of function call

Here's the attempt:

https://github.com/tomoyanonymous/mimium-rs/compare/dev...yutannihilation:syntactic-sugar-%40-tmp

tomoyanonymous:
I think it is not a problem to be a part of binary operators. In valid expression can be rejected at the type checking. Could you fix the precedence of the operator?

mimium-lang/src/compiler/parser.rs
```diff
@@ -116,11 +116,13 @@ where
...
         .then(op.then(prec).repeated())
         .foldl(move |x, ((op, opspan), y)| {
-            Expr::Apply(
-                Expr::Var(op.get_associated_fn_name()).into_id(opspan),
-                vec![x, y],
-            )
-            .into_id(x.to_span().start..y.to_span().end)
+            let arg = match op {
+                // A@B is a syntactic sugar of _mimium_schedule_at(B, A)
+                Op::At => vec![y, x],
```
tomoyanonymous, line 121: Is it better to fix the order of the argument of `_mimium_schedule_at` itself for the simplicity of here?

yutannihilation:
Sure.

Let me check if my understanding is correct. The lowest precedence means these two are the same?

```
updater@(now+1.0)

updater@now + 1.0
```

mimium-lang/src/compiler/parser.rs
yutannihilation: Thanks, I was wondering about this, too. Are there a chance that `@` eventually accept function call with arguments like described in [the v1 document](https://mimium.org/ja/docs/users-guide/language-specification/grammar-rule/#%E6%BC%94%E7%AE%97%E5%AD%90%E3%81%AB%E3%82%88%E3%82%8B%E9%81%85%E5%BB%B6%E5%AE%9F%E8%A1%8C)?

```
println(100)@0
```

If so, is `_mimium_schedule_at(0, println, 100)` better than `_mimium_schedule_at(println, 0, 100)` because the function and argument are in order?

tomoyanonymous:
I'd say yes... but is it perhaps confusing for user? Let me check how I implemented previous version of the compiler...

tomoyanonymous:
https://github.com/mimium-org/mimium/blob/dev/src/compiler/mimium.yy#L346-L351

I think `updater@now + 1.0` is allowed in the previous versions too.

Having infix operator that never return values, is itself something weird, but I keep this spec for now...

yutannihilation:
Thanks. I too think this won't be a real problem, just a matter of visual soundness; as `@` can be used only for a function that returns `()`, no binary operator should be valid.

Btw, I think it's probably common to specify a time that relative to `now`, so it might be useful to have some special syntax for `(now + ...)`. For example,

```
updater@+1.0
```

tomoyanonymous:
> `updater@+1.0`

That looks nice!!

PR #47, opened 2024-09-09
Support tuple assignment on global

yutannihilation:
Currently, tuple assignment doesn't work on global.

```
let (x, y) = (1.0, 2.0)
fn dsp(){
    x
}
```

```
thread 'main' panicked at mimium-lang\src\compiler\bytecodegen.rs:222:14:
value reg(9) not found
```

It seems this is because global value is created only for the whole tuple and not for the elements. I'm wondering if this can be supported by moving the creation of a global value to `add_bind_pattern()` so that it can handle the case of recursion.

### DIFF
--- a/mimium-lang/src/compiler/bytecodegen.rs
+++ b/mimium-lang/src/compiler/bytecodegen.rs
@@ -471,23 +471,24 @@ impl ByteCodeGenerator {
                 vec![VmInstruction::Closure(dst, idx)]
             }
             mir::Instruction::CloseUpValues(src, ty) => {
-                let addrs = if matches!(ty.to_type(), Type::Tuple(_) | Type::Array(_)) {
-                    // TODO
-                    &[self.find_keep_and_get_element(src, *ty, 0, 0)]
-                } else {
-                    &[(self.vregister.find_keep(src).unwrap(), 1)]
-                };
+                let flattened = ty.flatten();
+                let base_addr = self.vregister.find_keep(src).unwrap();
 
-                addrs
-                    .iter()
-                    .map(|(addr, tsize)| {
+                let mut offset = 0;
+                let mut instructions = Vec::with_capacity(flattened.len());
+                for elem_t in flattened {
+                    let tsize = Self::word_size_for_type(elem_t);
+                    if elem_t.to_type().is_function() {
                         self.vregister
                             .get_top()
                             .0
-                            .insert(dst.clone(), MemoryRegion(*addr, *tsize));
-                        VmInstruction::Close(*addr)
-                    })
-                    .collect::<Vec<_>>()
+                            .insert(dst.clone(), MemoryRegion(base_addr + offset, tsize));
+                        instructions.push(VmInstruction::Close(base_addr + offset))
+                    }
+                    offset += tsize;
+                }
+
+                instructions
             }
             mir::Instruction::GetUpValue(i, ty) => {
                 let upval = &mirfunc.upindexes[*i as usize];
--- a/mimium-lang/src/compiler/bytecodegen.rs
+++ b/mimium-lang/src/compiler/bytecodegen.rs
@@ -480,6 +480,7 @@ impl ByteCodeGenerator {
                 vec![VmInstruction::Closure(dst, idx)]
             }
             mir::Instruction::CloseUpValues(src, ty) => {
+                // src might contain multiple upvalues (e.g. tuple)
                 let flattened = ty.flatten();
                 let base_addr = self.vregister.find_keep(src).unwrap();
 
--- a/mimium-lang/src/compiler/bytecodegen.rs
+++ b/mimium-lang/src/compiler/bytecodegen.rs
@@ -176,14 +176,14 @@ impl ByteCodeGenerator {
         inst: F,
         dst: &Arc<mir::Value>,
         v1: &Arc<mir::Value>,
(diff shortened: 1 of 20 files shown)


yutannihilation:
It seems the current failure is not related to global things.

The problem is probably here. This emits `Instruction::CloseUpValue()` on the whole return type, not the actual closure element of it. So, in the case when the closure locates at the first element of the tuple, it succeeds because the addresses match. But, in other cases, it fails.

https://github.com/tomoyanonymous/mimium-rs/blob/9d5ae01d5b2386d9c2e8435df1f1c5b0f6f23e7e/mimium-lang/src/compiler/mirgen.rs#L607-L610

I think there are two ways to fix this, and I'm thinking which one is better.

* Add a field to `Instruction::CloseUpValue()` and store the offset calculated at the MIR stage.
* Inspect and close recursively at the bytecode stage

tomoyanonymous:
Ah. The return value may contain 2 or more closures and current implementation can not handle such cases indeed.

For example, if we have return value of  `(function, float, function)` type, we can emit bytecodes `Close n` and `Close n+2` respectively from the type information.

So I think we do not need to modify MIR definition though it is more clear to rename `CloseUpvalue` to `CloseUpvalue**s**`.

yutannihilation:
Thanks for the help!

> So I think we do not need to modify MIR definition

If I understand correctly, `ByteCodeGenerator` doesn't know the exact type of a value of `Value::Register`, so the MIR needs to be modified to include `TypeNodeId`. I think that's why `GetElement()` contains `TypeNodeId`. I'm not sure if this is ideal, but I think the current implementation works in this way.

> it is more clear to rename `CloseUpvalue` to `CloseUpvalues`.

This sounds a good idea!

tomoyanonymous:
> ByteCodeGenerator doesn't know the exact type of a value of Value::Register, so the MIR needs to be modified to include TypeNodeId.

Ah, that's right. We need to add type information tothe instruction...

yutannihilation:
Thanks, I'll explore in that direction.

yutannihilation:
Current status:

* The first example now works.
* I had to change the return type of `emit_instruction()` from `Option<VmInstruction>` to `Vec<VmInstruction>` because `CloseUpvalues` emits multiple `Close` instructions. Probably this can be pushed to the `Vec` directly.
* This new test doesn't work. It seems there's some mistake in my implementation when there's multiple closures in a tuple.

```
fn makecounter(){
    let x = 0.0
    let countup = | |{
        let res = x
        x = x+1.0
        res
    }

    let y = 0.0
    let countdown = | |{
        let res = y
        y = y-1.0
        res
    }
    (countup,0.0,countdown)
}
let (c1,_,c2) = makecounter();
fn dsp(){
    (c1(),c2())
}
```

mimium-lang/src/compiler/bytecodegen.rs
```diff
@@ -458,31 +458,47 @@ impl ByteCodeGenerator {
...
+
+                let mut offset = 0;
+                let mut instructions = Vec::with_capacity(flattened.len());
+                for elem_t in flattened {
+                    let tsize = Self::word_size_for_type(elem_t);
+                    if elem_t.to_type().is_function() {
+                        self.vregister
+                            .get_top()
+                            .0
+                            .insert(dst.clone(), MemoryRegion(base_addr + offset, tsize));
```
yutannihilation, line 495: It seems there are multiple problems.

One apparent problem (my mistake) is that,  `dst` is overwritten here when there are multiple closures.

However, even if the tuple contains one closure, this fails. So, this should be a different problem. I'm yet to figure out what's happening here.

```
fn makecounter(){
    let x = 0.0
    let countup = | |{
        let res = x
        x = x+1.0
        res
    }
    (0.0,0.0,countup)
}
let (_,_,c2) = makecounter()
fn dsp(){
    (c2(),0.0)
}
```
```
thread 'closure_counter_tuple' panicked at mimium-lang/src/runtime/vm.rs:421:14:
range end index 10 out of range for slice of length 9
```

tomoyanonymous: That code emit bytecode like this
```
makecounter
nparams:0 nret: 3
upindexes: []  state_size: 0  
constants:  [0, 2]
instructions:
  movc       1 0
  mov        0 1
  movc       2 1
  closure    2 2
  mov        1 2
  movc       5 0
  mov        2 5
  movc       5 0
  mov        3 5
  mov        5 1
  mov        4 5
  close      4
  ret        4 3
```

It closes and returns `4` but it should close `4` and return `2 3` right?

yutannihilation: Oh, thanks for catching! I think it should be. I'll check how that bytecode is generated.

yutannihilation:
While the problem related to upvalue is not solved yet, I decided narrow the scope of this pull request to what I thought at first; this pull request only fixes tuple assignment of global values like this.

```
let (x, y) = (1.0, 2.0)
fn dsp(){
    x
}
```

The upvalue related fix seems a bit broader than I expected, and might overlap with the ongoing pull request https://github.com/tomoyanonymous/mimium-rs/pull/50. So, while I don't give up yet, it's better to create another pull request.

tomoyanonymous:
okay! So i'll merge this first.

yutannihilation:
Thanks!

PR #48, opened 2024-09-09
Refactor parser to parse statement without return value

tomoyanonymous:
Continued from #41, #46. 
This PR introduces an intermediate struct `Statement` for the parser.
With this PR, single function call without return value including scheduling `a@b` can be correctly parsed.


mimium-lang/src/compiler/parser/test.rs
```diff
@@ -240,6 +247,57 @@ fn test_fndef() {
...
+                ],
+                None,
+                Expr::Var("input".to_symbol()).into_id(21..26),
+            )
+            .into_id(0..28),
+            None
+        ))
+    );
+    test_string!(
+        "fn hoge(input,gue){\n input\n}\nfn hoge(input,gue){\n input\n}",
```
tomoyanonymous, line 296: I have not given correct span for this test result but it passes somehow. `PartialEq` for `ExprNodeId` is comparing both content and span but maybe is it not working correctly?

mimium-lang/src/interner.rs
```diff
@@ -136,13 +136,13 @@ pub struct TypeNodeId(pub TypeKey);
 
 impl PartialEq for ExprNodeId {
     fn eq(&self, other: &Self) -> bool {
-        self.to_expr() == other.to_expr() && self.to_span() == self.to_span()
+        self.to_expr() == other.to_expr() && self.to_span() == other.to_span()
```
tomoyanonymous, line 139: Ok, it was comparing identical span....

yutannihilation: Aw, thanks for catching! This was my oversight in #27...

PR #50, opened 2024-09-10
Add `letrec` syntax

tomoyanonymous:
Currently, recursive function can only be defined through `fn hoge(args) { expr }` syntax, which is a syntactic sugar to `letrec hoge = |args| {expr}` though the `letrec` cannot be allowed to parse, and `fn` syntax cannot be nested.
In this PR, we add a direct expression of `letrec` syntax to allow defining a recursive function inside function.

```rust
fn makecounter(){
    let x = 0.0
    letrec gen = | |{
        x = x+1.0
        gen@now+1.0
    }
    gen@1.0
    let getter = | | {x}
    getter
}
let x_getter = makecounter();
fn dsp(){
    x_getter()
}
```

### DIFF
--- a/mimium-cli/tests/intergration_test.rs
+++ b/mimium-cli/tests/intergration_test.rs
@@ -221,6 +221,12 @@ fn closure_counter2() {
     assert_eq!(res, ans);
 }
 #[test]
+fn closure_escape_3nested() {
+    let res = run_file_test_mono("closure_escape_3nested.mmm", 5).unwrap();
+    let ans = vec![1.0, 2.0, 3.0, 4.0, 5.0];
+    assert_eq!(res, ans);
+}
+#[test]
 fn closure_counter_tuple() {
     let res = run_file_test_stereo("closure_counter_tuple.mmm", 5).unwrap();
     #[rustfmt::skip]
@@ -233,6 +239,7 @@ fn closure_counter_tuple() {
     ];
     assert_eq!(res, ans);
 }
+
 #[test]
 fn hof_state() {
     let res = run_file_test_mono("hof_state.mmm", 10).unwrap();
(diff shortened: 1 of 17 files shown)


tomoyanonymous:
Currently, upvalues are not shared between recursively generated closure instances...

```rust
fn makecounter(){
    let x = 0.0
    letrec gen = | |{ // <- when this closure is made, x is correctly found.
        x = x+1.0
        gen@now+1.0 // <- when the closure made by higher-order function call, x can not be found correctly.
    }
    gen@1.0
    let getter = | | {x}
    getter
}
let x_getter = makecounter();
fn dsp(){
    x_getter()
}
```

tomoyanonymous:
It is ready to merge.

Note that I disabled auto de-allocation of the closure introduced in #28 because the algorithm was imcomplete.
It can handle this temporal recursion code, 
```rust
fn makecounter(){
    let x = 0.0
    letrec gen = | |{
        x = x+1.0
        gen@now+1.0
    }
    gen@1.0
    let getter = | | {x}
    getter
}
let x_getter = makecounter();
fn dsp(){
    x_getter()
}
```

But in this indirect temporal recursion code, which is semantically equivalent (`scheduler_counter_indirect.mmm`), it drops inline closure`| |{gen()}` that should be kept until used in the scheduler.

```rust
fn makecounter(){
    let x = 0.0
    letrec gen = | |{
        x = x+1.0
       | |{gen()}@(now+1.0) <- this closure never escapes and dropped in the end of scope
    }
    | |{gen() }@1.0
    let getter = | | {x}
    getter
}
let x_getter = makecounter();
fn dsp(){
    x_getter()
}
```

I will try to solve this in another PR, basically with either of 2 ways.

- handle `_mimium_scheduler_at` as a special function 
- implement straight forward garbage collection (mark&sweep or reference counting)

I think we need to choose the latter, because it is harder to analyze nested closure escapement statically.
This test code (closure_escape_3nested.mmm) fails if auto de-allocation is enabled.
```rust
fn outer(){
  let x = 0.0
  let inner = | | { 
    x = x+1.0
    x
  } 
  let mid = | | {inner}
  mid
}
let mid = outer();
let inner = mid();
fn dsp(){
 let r = inner()
 r
}
```

PR #51, opened 2024-09-10
Fix CloseUpValues on a tuple containing closures

yutannihilation:
(It seems this conflicts with #50, so I'll recreate this after #50 gets merged.)

Continued from https://github.com/tomoyanonymous/mimium-rs/pull/47#issuecomment-2339713677

### DIFF
--- a/mimium-cli/tests/intergration_test.rs
+++ b/mimium-cli/tests/intergration_test.rs
@@ -220,19 +220,19 @@ fn closure_counter2() {
     let ans = vec![0.0, 1.0, 2.0, 3.0, 4.0];
     assert_eq!(res, ans);
 }
-// #[test]
-// fn closure_counter_tuple() {
-//     let res = run_file_test_stereo("closure_counter_tuple.mmm", 5).unwrap();
-//     #[rustfmt::skip]
-//     let ans = vec![
-//         0.0,  0.0,
-//         1.0, -1.0,
-//         2.0, -2.0,
-//         3.0, -3.0,
-//         4.0, -4.0
-//     ];
-//     assert_eq!(res, ans);
-// }
+#[test]
+fn closure_counter_tuple() {
+    let res = run_file_test_stereo("closure_counter_tuple.mmm", 5).unwrap();
+    #[rustfmt::skip]
+    let ans = vec![
+        0.0,  0.0,
+        1.0, -1.0,
+        2.0, -2.0,
+        3.0, -3.0,
+        4.0, -4.0
+    ];
+    assert_eq!(res, ans);
+}
 #[test]
 fn hof_state() {
     let res = run_file_test_mono("hof_state.mmm", 10).unwrap();
--- a/mimium-cli/tests/intergration_test.rs
+++ b/mimium-cli/tests/intergration_test.rs
@@ -222,7 +222,6 @@ fn closure_counter2() {
 }
 #[test]
 fn closure_counter_tuple() {
-    let res = run_file_test_stereo("closure_counter_tuple.mmm", 5).unwrap();
     #[rustfmt::skip]
     let ans = vec![
         0.0,  0.0,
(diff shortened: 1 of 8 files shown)


yutannihilation:
Is this supposed to work at the moment? Or, is it that this requires some missing feature? Sorry, I might not catch up with the current development status...

``` diff
 // counter implemented using closure. for the test of assignment expression
 fn makecounter(){
     let x = 0.0
     let countup = | |{
         let res = x
         x = x+1.0
         res 
     }
     countup
 }
-let myc = makecounter();
 fn dsp(){
+    let myc = makecounter()
     myc()
 }
```

yutannihilation:
Ah, no, apparently, this doesn't work because `myc` is freshly created on every cycle. My bad...

yutannihilation:
While I think this probably won't be merged before #50, this should be basically ready for review now.

I tried to change `emit_instruction()` to push instructions directly to `funcproto` or `bytecode_dst` in general, but I couldn't make it in a nice code. So, I decided to modify the `CloseUpValues` case only.

While everything seems to work fine, I'm still struggling to understand why it doesn't make any problem to overwrite the same `dst` in this `insert()`. I thought each element requires a unique register...

https://github.com/tomoyanonymous/mimium-rs/blob/2837c2d959997859de4299f9bf0ef2ad828f1765/mimium-lang/src/compiler/bytecodegen.rs#L489-L499

tomoyanonymous:
>While everything seems to work fine, I'm still struggling to understand why it doesn't make any problem to overwrite the same dst in this insert(). I thought each element requires a unique register...

This is probably because of the meaningless spec of the `CloseUpvalues` instruction that always return same address as the source address. This is because MIR follows Single Static Assignment principle, (register must be consumed once at most) except for the registers made with `alloc` operation.

I realized that, in MIR generator, return value made with `CloseUpvalues` is sometimes referred and sometimes discarded.

```rust
                        (Value::Function(i), _) => {
                            let idx = ctx.push_inst(Instruction::Uinteger(*i as u64));
                            let cls = ctx.push_inst(Instruction::Closure(idx));
                            let newres = ctx.push_inst(Instruction::CloseUpValues(cls.clone(), rt));//used
                            let _ = ctx.push_inst(Instruction::Return(newres, rt));
                        }
                        (_, _) => {
                            if rt.to_type().contains_function() {
                                let _ = ctx.push_inst(Instruction::CloseUpValues(res.clone(), rt));//discarded
                                let _ = ctx.push_inst(Instruction::Return(res.clone(), rt));
                            } else {
                                let _ = ctx.push_inst(Instruction::Return(res.clone(), rt));
                            }
                        }
```

tomoyanonymous:
The code in bytecodegen should be like this?

```rust
            mir::Instruction::CloseUpValues(src, ty) => {
                // src might contain multiple upvalues (e.g. tuple)
                let flattened = ty.flatten();
                let base_addr = self.vregister.find_keep(src).unwrap();
                let mut offset = 0;
                self.vregister
                    .get_top()
                    .0
                    .insert(dst.clone(), MemoryRegion(base_addr, Self::word_size_for_type(*ty)));
                for elem_t in flattened {
                    let tsize = Self::word_size_for_type(elem_t);
                    if elem_t.to_type().is_function() {
                        bytecodes_dst.push(VmInstruction::Close(base_addr + offset))
                    }
                    offset += tsize;
                }
                None
            }
```
I'm thinking to not using return value of `CloseUpvalues` in MIR like `Store`, `SetGlobal`, and `SetUpValue` because the source of the CloseUpvalues is Closure and thus it has specific memory address. It does not matter the value is referred multiple times.

yutannihilation:
> This is probably because of the meaningless spec of the `CloseUpvalues` instruction that always return same address as the source address.

> I realized that, in MIR generator, return value made with `CloseUpvalues` is sometimes referred and sometimes discarded.

Thanks! I was wondering when the spec is meaningful. So, it seems we can just ignore it consistently.

tomoyanonymous:
Okay #50 may includes bigger change in MIR funcproto. So I'll merge this first. Thanks!

yutannihilation:
I see. I'll keep an eye on #50. Thanks for merging!

PR #52, opened 2024-09-12
Improve type mismatch error

yutannihilation:
This pull request is to improve error messages of `ErrorKind::TypeMismatch`. The current error message is a bit confusing because the span of the actual error location is not propagated.

**Code:**

```
let unrelated = 0.0

fn foo() -> float{
    1.0
}
fn dsp(){
    let (x, y) = foo()
    x
}
```

**Error message:**

```
   ╭─[../path/to/tmp.mmm:7:18]
   │
 1 │ let unrelated = 0.0
   · │
   · ╰─ Type Mismatch, between number and (i(3()),i(4()))
───╯
```

This pull request improves this *to some extent* by the following changes:

1. Explicitly propagate the appropriate `span`.
2. Improve the parser to include the span of `( )` in `Apply`.
3. Show user friendly message like `(?,?)` instead of `(i(3()),i(4()))`.

The error message will look like this:

```
   ╭─[../path/to/tmp.mmm:7:18]
   │
 7 │     let (x, y) = foo()
   ·                  ──┬──
   ·                    ╰──── Type Mismatch, between number and (?,?)
───╯
```

---

Some thoughts:

* This pull request is not intended to be complete. This addresses only a few of the cases. I hope we can improve the error message step by step.
* Probably, `Type::into_id()` should take `span`. It's hard to track where the span is replaced with `0..0`. It might be hard to figure out what span is appropriate, but I'm feeling any span is better than `0..0`.
* The explicit propagation was needed because the `Span` corresponding to a `Type` might not be appropriate for the error message. For example, when I attempted to store the `Span` and use it, I got the following error message. Apparently, `foo()`'s type is determined by the `-> float` annotation, but we expect `let (x, y) = foo()` is where the error happens.

```
   ╭─[../path/to/tmp.mmm:7:18]
   │
 3 │ fn foo() -> float{
   ·             ──┬──
   ·               ╰──── Type Mismatch, between number and (i(3()),i(4()))
───╯
```

### DIFF
--- a/mimium-lang/src/compiler/parser.rs
+++ b/mimium-lang/src/compiler/parser.rs
@@ -232,16 +232,12 @@ fn atom_parser<'a>(
 }
 fn expr_parser(expr_group: ExprParser<'_>) -> ExprParser<'_> {
     recursive(|expr: Recursive<Token, ExprNodeId, Simple<Token>>| {
-        let parenitems =
-            items_parser(expr.clone()).delimited_by(just(Token::ParenBegin), just(Token::ParenEnd));
-        let folder = |f: ExprNodeId, args: Vec<ExprNodeId>| {
+        let parenitems = items_parser(expr.clone())
+            .delimited_by(just(Token::ParenBegin), just(Token::ParenEnd))
+            .map_with_span(|args, args_span| (args, args_span));
+        let folder = |f: ExprNodeId, (args, args_span): (Vec<ExprNodeId>, Span)| {
             let f_span = f.to_span();
-            // TODO: this doesn't include the span of the closing parenthesis
-            let span_end = match args.as_slice() {
-                [.., end] => end.to_span().end,
-                _ => f_span.end,
-            };
-            let span = f_span.start..span_end;
+            let span = f_span.start..args_span.end;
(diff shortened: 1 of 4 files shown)


tomoyanonymous:
The improvement looks good!
Ideally, we need to modify `ReportableError` trait so that it can have 0 or more spans(Ariadne crate can report the error caused by multiple locations [like main example image](https://github.com/zesterer/ariadne/).)

About the type unification, `unify_types` do not need to return `dyn ReportableError` directly but, for example, it is better to return another custom type `UnificationError(pub TypeNodeId, pub TypeNodeId)` and translate them in `infer_type` by giving spans with `map_err` into `InferError::TypeMismatch(TypeNodeId,Span,TypenodeId,Span)` that implements `ReportableError`. Maybe the code becomes a bit longer but it does not rely on span attached to `TypeNodeId`.

yutannihilation:
> `unify_types` do not need to return `dyn ReportableError` directly but, for example, it is better to return another custom type `UnificationError(pub TypeNodeId, pub TypeNodeId)`

Thanks for the suggestion, this sounds good to me!

PR #53, opened 2024-09-16
Use a custom error instead of passing span to `unify_types()`

yutannihilation:
This pull request is an attempt to implement what was suggested in https://github.com/tomoyanonymous/mimium-rs/pull/52#issuecomment-2348141627:

> About the type unification, `unify_types` do not need to return `dyn ReportableError` directly but, for example, it is better to return another custom type `UnificationError(pub TypeNodeId, pub TypeNodeId)` and translate them in `infer_type` by giving spans with `map_err` into `InferError::TypeMismatch(TypeNodeId,Span,TypenodeId,Span)` that implements `ReportableError`. Maybe the code becomes a bit longer but it does not rely on span attached to `TypeNodeId`.

(I'm not very familiar with how these errors should be organized, so I might need to go back and forth several times.)

### DIFF
--- a/mimium-lang/src/compiler/typing.rs
+++ b/mimium-lang/src/compiler/typing.rs
@@ -81,6 +81,27 @@ impl ReportableError for Error {
 }
 
 #[derive(Clone, Debug, PartialEq)]
+pub struct UnificationError(pub Type, pub Type);
+
+impl std::error::Error for UnificationError {}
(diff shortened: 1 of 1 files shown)


yutannihilation:
Closing for now. Maybe I can retry after I figure out a bit more about how the error should be organized.

PR #58, opened 2024-09-18
Implement Pipe

### LINKED ISSUE #56: Pipe operator
I [experimentally implemented a pipe operator,](https://github.com/tomoyanonymous/mimium-rs/compare/main...yutannihilation:mimium-rs:feat-pipe?expand=1) and got to know there are two types of pipes.

cf. https://github.com/tc39/proposal-pipeline-operator?tab=readme-ov-file#why-the-hack-pipe-operator

* explicit function call (Hack pipe): `value |> foo(%)`
* implicit function call (F# pipe): `value |> foo`

It seems my implementation is a simpler version of "Hack pipe," which is without the placeholder `%` by assuming the value is inserted into the first argument. Since my main language is R, I naturally assumed this is the pipe operator. On the other hand, Mimium's pipe is "F# pipe". I think this needs some discussion because both pipe have their pros and cons.

Is it common to use the F# style in this area? If I understand correctly, F#'s pipe is designed for combination with partial application. I'm wondering if there's any plan.

(I'm asking this simply because it looks easy to implement. Not sure about the importance of the feature itself)

yutannihilation:
Close #56

- [x] Support placeholder
- [x] Support pipe
- Allow an expression with a binary op to be multiline
    - [x] trailing binary op (`value |> \n f(_)`)
    - [x] leading binary op (`value \n |> f(_)`)
- ~~Support passing a tuple of parameters~~ (out of scope of this pull request)


tomoyanonymous:
Why dont we convert the placeholder after all the parsing finished (like `self_convert.rs`)?

I want to support pipe without placeholder if the number of argument is 1, because it keeps the semantics simple (`a |> b` is always equivalent to `b(a)`).

```rust
fn dsp(){
    1.0 |>
        do_plus(_, 3.0) |> 
        do_div(16.0, _) |>
        //sqrt(_) //not like this
        sqrt // but like this
        
}
```

If we could implement as a separated process,  the placeholder can also be used with binary operator like `(_ + x)` and I think it is desirable.

yutannihilation:
Thanks, there are several things I still don't figure out.

> I want to support pipe without placeholder if the number of argument is 1

If I understand correctly, in semantics (discuss later), this is incompatible with the support of tuple. I'm not 100% sure, but I think supporting both would make the type inference ambiguous. In this case, the type of `some_fun` can be either `(number) -> ?` or `((number,)) -> ?`.

```
1.0 |> some_fun
```
```
(1.0, ) |> some_fun
```

In this case, it's probably possible to treat the case when LHS is a tuple expression specially, but if it's a variable containing tuple, it's hard to determine.

```
(1.0, ) |> some_fun
```
```
let x = (1.0, )
x |> some_fun
```

I said "semantically," but `1.0` and `(1.0, )` probably have the same representation on stack, so it won't make problems on execution. Only the type inference (and the impact of the result on other inferences) matters.

> If we could implement as a separated process, the placeholder can also be used with binary operator like `(_ + x)` and I think it is desirable.

This is a good point. I assumed `_` is allowed only as a direct argument of a function call. e.g. `value |> g(f(_))` is not supported in my current implementation.

tomoyanonymous:
I think the parameter pack is a special kind of a type coercion from tuple type, though I don't know if there is theoritical discussion.

> In this case, the type of some_fun can be either (number) -> ? or ((number,)) -> ?.

I guess the most of inference of the function type is done in the evaluation of function body, not in the application. 

Basically we need to modify type inference code and mir generator code.What I imagine as pseudo-code in type inference and MIR generator is like this.

```rust
Expr::Apply(f, args)=>{
   let ft = infer_type(f)?;
   let  argtypes = infer_vec(args)?;
   match (ft.to_type(), argtypes.as_slice())=>{
      (Type::Function(at,rt), args)  =>{
         self.unify_vec(at,argtypes)?;
         Ok(rt)
     }
      (Type::Function(at,rt), [Type::Tuple(t_argtypes)]) if at.len() == t_argtypes.len() =>{
         self.unify_vec(at,t_argtypes)?;
         Ok(rt)
    }
    _=>{
     // type mismatch
    }
  }
}
```

```rust
Expr::Apply(f, args)=>{
   let (freg,ft) = eval(f)?;
   let (ares, argtypes) = eval_args(args)?;
   match (ft.to_type(), argtypes.as_slice())=>{
      (Type::Function(at,rt), args) if at == argtypes =>{
       // normal mir generation
     }
      (Type::Function(at,rt), [Type::Tuple(t_argtypes)]) if at == t_argtypes =>{
        let aregs = ares.iter().map(|a| self.push_inst(Instruction::GetElement(...)) ).collect::<Vec<_>>();
        ...
    }
    _=>{
     // type mismatch
    }
  }
}
```

When the generic type inference is introduced in #59, maybe the type inference for single tuple application may become ambigious, but it's not the problem because the allocation layout will not change between single element tuple and its content like you said.

>  I assumed `_` is allowed only as a direct argument of a function call. e.g. `value |> g(f(_))` is not supported in my current implementation.

Note that this won't cause parse error, but it will be expanded into `value |> g(|x| {f(x)})`, which causes a type mismatch error

yutannihilation:
> I guess the most of inference of the function type is done in the evaluation of function body, not in the application.

I agree the most of the cases, it doesn't matter. But, while I cannot come up with a real problem, it might matter in some cases like the following code; `x`'s type is inferred from the function body of `foo()` and the application of `bar()`.

```
let x = foo()
x |>
    bar
```

Maybe this is just because I'm too afraid of ambiguities. I'll leave the parameter pack cases later (or a separate pull request).


> Note that this won't cause parse error, but it will be expanded into `value |> g(|x| {f(x)})`, which causes a type mismatch error

Oh, this is not what I expected. Thanks for clarifying!

yutannihilation:
> If we could implement as a separated process, the placeholder can also be used with binary operator like `(_ + x)` and I think it is desirable.

> `value |> g(|x| {f(x)})`

Okay, I think I understand the semantics at last. I thought

* `_` can be used only on the RHS of `|>`
* `|>` takes either a function call (`f(a, b, _)`) or a single-argument function (`f`)

But, the actual rule is,

* `_` can be used anywhere and it creates a single-argument function (e.g. `let f = _ + 1`  is `let f = |x| x + 1`)
* `|>` always takes a function, not a function call

right?

tomoyanonymous:
> - _ can be used anywhere and it creates a single-argument function (e.g. `let f = _ + 1` is `let f = |x| x + 1`)
> - `|>` always takes a function, not a function call

Yes, specifically, if the AST matched to the specific pattern, translate it to new lambda expr like this. 

```rust
Expr::Apply(f,args) if args.iter().any(|e| matches!(e,Expr::Placeholder)){
   let new_args = args.iter()
                                  .filter(|e| matches!(e,Expr::Placeholder))
                                  .enumerate()
                                  .map(|(i,_)| format!("arg_{i}"))
                                  .collect::<Vec<_>>();
   let typedids = ...;
   let rvars = ...;
   Expr::Lambda(typedids , Expr::Apply(f, rvars))
}
```

This desugaring process do not need to special care about the pipe operator.

tomoyanonymous:
> I'll leave the parameter pack cases later (or a separate pull request)

yeah, the parameter pack can be implemented separately.

yutannihilation:
Got it, thanks!

yutannihilation:
The implementation is mostly done. One last thing is to allow placing `|>` at the head of the line.

I'm afraid leading binary operator is tricky because of unary `+` and `-`. For example, the return value of this function cannot  be syntactically determined; it can be `-x` or `bar() - x`.

```
fn foo() {
  bar()
  -x
}
```

So, I think this pull request is going to deal with this by treating `|>` specially than other binary infix operators. But, I'm open to any suggestions.

tomoyanonymous:
> So, I think this pull request is going to deal with this by treating |> specially than other binary infix operators.

I think this is the good choice! I'll take a look later. Thanks!

PR #59, opened 2024-09-18
Generic type inference(WIP)

tomoyanonymous:
This is the experimental implementation of the type inference with generics.

The implementation is based on Level-based Hindley-Milner type inference algorithm.

https://rhysd.hatenablog.com/entry/2017/12/16/002048

Note that currently only the let-polymorphism is supported, not the complete parametric polymorphism(Type declarations with type parameters are not allowed for now.)

### DIFF
--- a/mimium-lang/src/compiler/bytecodegen.rs
+++ b/mimium-lang/src/compiler/bytecodegen.rs
@@ -127,7 +127,7 @@ impl ByteCodeGenerator {
             Type::Function(_, _, _) => 1,
             Type::Ref(_) => 1,
             Type::Code(_) => todo!(),
-            Type::TypeScheme(_, _) => todo!(),
+            Type::TypeScheme{..} => unreachable!(),
             Type::Intermediate(_) => 1, // TODO
             Type::Unknown => todo!(),
         }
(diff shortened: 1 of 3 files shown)


tomoyanonymous:
I'm going to merge this for now because the monomorphization code is gonna be a bigger change, and current improvement does not break existing tests. 
I will continue this task after completed audio file player implementation.

yutannihilation:
Just curious. In what case does mimium semantically require integer? I was thinking array indexing needs an integer index, but I just found the automatic interpolation feature allows float. Likewise, `@` can probably accept float. Array assignment? Or, is it mainly for performance (integer arithmetic vs floating-point arithmetic)?

tomoyanonymous:
> Or, is it mainly for performance (integer arithmetic vs floating-point arithmetic)?
Mostly yes. Array interpolation should not be used if the indice are integer type.

Another reason is, some process like pseudo-random generator requires bitwise operations like bitshift for xorshift.
Faust has  `int` operator as a explicit cast but bitshift operation can also be used, and I guess it relies on the target language's implicit cast feature (not confident though).

If the generics can be introduced, the user do not need to care about whether the number is int or float in the most cases because the implicit type coercion from int to float is done, but I have not decided whether the implicit coversion from float to int are allowed, because allowing the bidirectional conversion may make the implementation a bit complex. Maybe all the number literals should be interpreted as float by default, and the special suffix like `42i` should be added.(`i` suffix may be confusing between imaginary number though)

yutannihilation:
I see, thanks for the details! Bitwise operations definitely require integer.

> Maybe all the number literals should be interpreted as float by default, and the special suffix like `42i` should be added.

I agree with this strategy.

PR #60, opened 2024-09-21
chore: update ariadne to 0.4.1

yutannihilation:
For some reason, mimium uses a bit outdated version of ariadne. It seems there's no significant change in the changelog, so this isn't something very urgent. Mostly just for house keeping.

cf. https://github.com/zesterer/ariadne/blob/main/CHANGELOG.md


tomoyanonymous:
Thanks! I think it was not updated since I added to the repository...

PR #62, opened 2024-09-21
Add CLI an option to generate values in CSV format

yutannihilation:
This pull request adds several options to write the generated values to stdout as CSV.

* `--output-format`: Specify the output format of file.
* `--output`/`-o`: Specify the path to the output CSV file. If not specified, stdout is used.
* `--times`: How many times the CLI should execute the code (default: 10)

```
fn phasor(freq){
    (self + 6.28/freq)%1.0
}
fn sinwave(freq){
    sin(phasor(freq)) + sin(phasor(freq*2.0))
}
fn dsp(){
    (sinwave(440.0),sinwave(220.0),sinwave(110.0))
}
```

``` console
❯ cargo run -- --output-format=csv tmp.mmm 2>$null
ch0,ch1,ch2
0.0,0.0,0.0
0.021408545756451732,0.04281382071922544,0.08560147875345994
0.04281382071922544,0.08560147875345994,0.17099377325384515
0.06421255470540224,0.1283368309581146,0.25596832404317227
0.08560147875345994,0.17099377325384515,0.34031781926405735
0.10697732573366744,0.21354626012028866,0.4238368134925043
0.12833683095811457,0.2559683240431722,0.5063223406277034
0.1496767327902555,0.2982340948993103,0.587574518887827
0.17099377325384513,0.34031781926405735,0.6673971459863752
0.19228469864114656,0.3821938796258443,0.7455982825954395
```

In order to enable this, this pull request made some changes on `Driver`-related things.

* Rename `MockDriver` to `LocalBufferDriver` because this is not only for tests now.
  * Instead of `play_times()`, it now sets `times` in `new()` and `play()`, then expose the result by `get_generated_samples()`.
  * Like `CsvDriver`, a driver can wrap this struct and read `get_generated_samples()`. Probably, this can be more efficient by inventing an interface that `LocalBufferDriver` directly writes to the destination, but I think this should be enough for now.
* Move `buffer_size` arg from `init()` to `new()`. By doing this, CLI now can handle only `Box<dyn Driver>`.
  * `new()` is out of `Driver` trait so that each struct can have their unique parameters.
  * `init()` can take only common parameters.

This also contains some minor improvements:

- Make options mutually exclusive
- Exit with error after report()


yutannihilation:
Btw, the motivation of this pull request is that I found this `sinwave.mmm` doesn't sound like sine wave and wanted to debug what's happening. Is this expected? Or, is this something wrong with my setup...?

https://github.com/tomoyanonymous/mimium-rs/blob/main/mimium-cli/tests/mmm/sinwave.mmm

Here's the plot using the data generated by this pull request on the mmm file:

image

tomoyanonymous:
`tests/mmm/sinwave.mmm` looks very old and unused, it has incorrect definition in the `phasor` function.
This is the correct definition(see examples/sinewave.mmm, that it the right implementation)

```
let pi = 3.14159265359
let sr = 44100.0
fn phasor(freq){
 ( (self + freq/sr)%1.0 ) *2.0 * pi
}
```

yutannihilation:
Ah, thanks! I was wondering why there are two versions of sInwave.mmm. Then, I'll remove it.

tomoyanonymous:
Anyway, the CSV output looks nice!

On the command line option, I think it is ok to merge it for now but I have some suggestion, Would we add something like `--length`,  `--output <file> (-o <file>)` and `--output-format(-of)`?
`--output` and `--output-format` are like the option in [pandoc](https://pandoc.org/MANUAL.html#general-options), if we set like `--output=plot.csv`, `--output-format=csv` is implicitly set and if we set `--output-format=csv` and `--output` is not specified, the CSV is written into stdout. With these options, we can split *how we write the output* and *where we write them into*.

This option can be extended to audio file export mechanism. (Writing audio file into stdout looks useless but it can be passed to ffmpeg without temporary file, for example.)

yutannihilation:
Sounds a great idea! Ill try it.

yutannihilation:
I think this is ready for review now. I updated the description above.

One minor caveat is that, unlike `pandoc`, a short option can only have one character, so `-of` is not allowed by clap.

> `--output <file> (-o <file>)` and `--output-format(-of)`?

------

Write to stdout

```
❯ cargo run -- --output-format=csv tmp.mmm 2>$null  
ch0
0.0
0.0626483241787478
0.12505253769702582
0.186967423254676
0.24814964662636757
0.30835871330736814
0.3673579142781011
0.424915257166559
0.48080437914347096
0.5348054379553847
```

Specify `--times`.

```
❯ cargo run -- --times 3 --output-format=csv tmp.mmm 2>$null
ch0
0.0
0.0626483241787478
0.12505253769702582
```

Write to a specified CSV file

```
❯ cargo run -- -o tmp.csv tmp.mmm 2>$null

❯ cat tmp.csv
ch0
0.0
0.0626483241787478
0.12505253769702582
0.186967423254676
0.24814964662636757
0.30835871330736814
0.3673579142781011
0.424915257166559
0.48080437914347096
0.5348054379553847
```

Write to a file whose extension is not `.csv`

```
❯ cargo run -- -o tmp tmp.mmm

...snip...

thread 'main' panicked at mimium-cli\src/main.rs:129:22:
cannot determine the output file format
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `target\debug\mimium-cli.exe -o tmp tmp.mmm` (exit code: 101)
```

Write to a file whose extension is not `.csv`, but with `--output-format` specified.

```
❯ cargo run -- -o tmp --output-format=csv tmp.mmm 2>$null

❯ cat tmp
ch0
0.0
0.0626483241787478
0.12505253769702582
0.186967423254676
0.24814964662636757
0.30835871330736814
0.3673579142781011
0.424915257166559
0.48080437914347096
0.5348054379553847

```

tomoyanonymous:
Sorry to be late! It looks okay. Thanks for the refinement!

yutannihilation:
Thanks for merging!

Issue #63, opened 2024-09-26
Cannot parse nested `if`


yutannihilation:
For some reason, this code cannot be parsed.

```
fn foo(x) {
  if (x == 0.0) {
    0.0
  } else {
    if (x > 0.0) {
      1.0
    } else {
      -1.0
    }
  }
}
```

To my eyes, the implementation looks correct. If the `expr_group` parser can parse `if (...) { ...`, it should also able to parse `if (...) { if (...) { ...` recursively. But...

https://github.com/tomoyanonymous/mimium-rs/blob/5c59926c6a4082970a6c0a9a2eae27aa8890c67b/mimium-lang/src/compiler/parser.rs#L339-L349

yutannihilation:
This works. So, it seems the problem is something related to handling of `{`.

```
fn foo(x) {
  if (x == 0.0) 0.0 else if (x > 0.0) 1.0 else -1.0
}
```

yutannihilation:
Oh, I see, thanks. I think this is a minor issue, so you can concentrate on implementing array. We can fix this later.

> I'm still struggling how not to cause an infinite loop in chumsky.

I see...

yutannihilation:
One more problem I'm investigating is that nested `if` seems to cause some stack inconsistency. I'm not sure if this is reproducible on other platform than my powerless Linux laptop. This is also not a very high priority. Some other change might eventually fix this. But, let me report here before I forget.

```
fn phasor(freq){
  (self + freq)
}
fn if_(x) {
  if (1.0) {
    -1.0
  } else {
    1.0
  }
}
fn foo(freq){
  let p = phasor(freq)
  if (1.0) {
    0.0
  } else {
    if_(p)
  }
}

fn dsp(){
  foo(440.0)
}
```

```
double free or corruption (out)
Aborted (core dumped)
```

tomoyanonymous:
The nested if with block can be parsed just by changing this from

```rust
 let block = block_parser(expr.clone()); 
```

to this.(It's fixed in #64)

```rust
 let block = block_parser(expr_group.clone()); 
```

yutannihilation:
Confirmed the first code now can be parsed with #64! (bytecodegen fails for some reason) Thanks.

tomoyanonymous:
[...]

PR #65, opened 2024-09-27
Handle nested `if`

yutannihilation:
This pull request is to show a possible fix for #63. (Since this requires the fix about parsing, the base branch is `feat-symphonia`, but this is not intended to be merged into the branch.)

I'm still not sure which is the best place to add `Jmp` to the `then` and `else` branches, but this pull request chooses bytecode instead of MIR. The main changes are

* record the index of the `phi` block in `JmpIf`
* Re-insert the destination of `Jmp` after the register is determined
* Push bytecodes to `bytecode_dst` to handle nested cases (but the code is messy...)


yutannihilation:
It seems most of the things work fine, but currently, the order of bytecode looks wierd, so probably I'm doing something wrong.

```
fn foo(x) {
  if (x == 0.0) {
    0.0
  } else {
    if (x > 0.0) {
      1.0
    } else {
      -1.0
    }
  }
}
```

```
foo
nparams:1 nret: 1
upindexes: []  state_size: 0
constants:  [0, 4607182418800017408]
instructions:
  mov        1 0
  movc       2 0
  eq         1 1 2   <----- probably a condition for outer if
  jmpifneg   2 4   <----- probably inner if?
  movc       2 1
  mov        4 2
  jmp        4
  movc       3 1
  negf       3 3
  mov        4 3
  jmpifneg   1 4     <----- outer if?
  movc       1 0
  mov        5 1
  jmp        5
  mov        2 0
  movc       3 0
  gt         2 2 3     <----- condition for inner if?
  mov        5 4
  ret        5 1
```

yutannihilation:
Since this pull request is based on `feat-symphonia`, this got automatically closed. I'll check if this still works on the current dev branch and reopen if possible (but I need some time to come back here...).

PR #67, opened 2024-10-14
Handle nested `if`

yutannihilation:
This is a rebased version of #65. 

It seems this still works on the current dev branch. I think the basic idea is done. What I was struggling to figure out is how to handle `funcproto.bytecodes` and `bytecodes_dst` better. Considering this pattern appears several times in `emit_instruction()`, probably this can be handled better on the caller's side, but I couldn't come up with a nice idea. Other than this point, I think this pull request is ready for review.

```rust
```

The description from #65:

I'm still not sure which is the best place to add `Jmp` to the `then` and `else` branches, but this pull request chooses bytecode instead of MIR. The main changes are

* record the index of the `phi` block in `JmpIf`
* Re-insert the destination of `Jmp` after the register is determined
* Push bytecodes to `bytecode_dst` to handle nested cases (but the code is messy...)


tomoyanonymous:
Thanks for the update! It seems okay.

> What I was struggling to figure out is how to handle funcproto.bytecodes and bytecodes_dst better.

Yeah, mutating both  `funcproto` and `bytecodes_dst` from the arguments at the same time is definitely a bad design... 

> record the index of the phi block in `JmpIf`
 
I think this is the vest choice in our language because the if always have some value(including unit type value). The notion of basicblock is coming from LLVM and other compiler infastructure design but it is more useful if the language have some loop construct in addition to the conditional. I won't be add such construct in the language primitive level, so having `Phi` instruction itself maybe overkill solution...

yutannihilation:
Thanks for the details, I was wondering about the design around `Phi`. It helps.

PR #68, opened 2024-10-16
Introduce Plugin System

tomoyanonymous:
This PR introduces some new interface to decouple scheduler and external functions from core functionality. I attach the brief description from `plugin.rs`.

---

## Plugin System for mimium
In order to extend mimium's capability to communicate between host system, mimium has its own FFI system.
The FFI is done through some traits in this plugin module in order to decouple dependencies(modules may depends on external crates).

There are 3 types of interfaces you need to define depending on what you need.

1. **IO Plugins** Sets of instance-free external functions such as `print` and `println`. They are mostly for glue functions for host system's IO. ~~`mimium-core` is an example of this type of module.~~ -> builtin_fn will be split in the future 
2. **External Unit Generator(UGen) Plugin.** If you need to define native Unit Generator, use `UGenPlugin` interface. In mimium code, you need to call higher-order function that returns instance of the UGen. You need to write small wrapper for this which simply calls object's constructor (In the future, this wrapper will be automatically implemented through proc-macro). Multiple instances may exist at the same time. `mimium-symphonia` is a example of this type of module.
3. **System Plugin**. If your plugin needs to mutate states of system-wide instance (1 plugin instance per 1 vm), you need to implement `SystemPlugin` traits. System plugin can have callbacks invoked at the important timings of the system like `on_init`, `on_sample` & so on. Internal synchronous event scheduler(mimium-scheduler) is implemented through this plugins system.  

---

Because the `SystemPlugin` need to be accessed mutably both from audio driver(e.g. `execute_task`) and mimium vm through user code(e.g. `schedule_at`), it creates double mutable borrow inevitably(perhaps this is my fault of interface design). Currently it breaks the double borrow checker with `UnsafeCell` and small unsafe codes in `to_ext_cls_info` function and callback invocation in `Driver`. (Use of `unsafe` in driver should be encapsulated within SystemPlugin code in the future...)

Also, Plugin and SystemPlugin use a dynamic dispatch through trait objects, runtime performance of scheduler function will be inferior to prior hard-coded implementation.

There are some significant change in VM implementation, `CallExtCls` operation is removed and the distinction between rust function pointer or rust closure is resolved at runtime now.

This is because it was hard to statically determine whether the external function is function pointer or rust closure at the type cheking or program generation stage (it would be possible but makes compiler's design complex).

I first tried to resolve it  when `link_function` called by replacing all `CallExtFun` operation that calls rust closure into `CallExtCls`, however, current implementation of the bytecode of `CallExtFun` is not the constant index of external functions but the register position which will contain the constant index, so the `MoveConst` operation also needed to be modified at the same time. I may change the first argument of `CallExtFun` to absolute index in the future, though it limits the number of external functions in the program up to 255.

## Needed Followups:

- Initialization order of `ExecContext `and `Driver` is messy, need to prepare some builder struct for these constructs.
- Builtin functions like `print`, `probe`, `println`, `probeln` can now be moved into separated crates.
- Need to prepare more concrete example. I am considering to prepare this by implementing random number generator plugin. It will be the combination of Native UGen plugin and System Plugin because the indeterminant random number generator (without given seed by user code) need to communicate host system and mutate system-wide instance variable.


tomoyanonymous:
Result of benchmark (compares `scheduler_counter.mmm`)

### w/o plugin system(before)

```
test scheduler::scheduler_counter ... bench:     388,219.26 ns/iter (+/- 45,024.05)
```

### w/ plugin system(after)
```
test tests::scheduler_counter ...    bench:     385,839.58 ns/iter (+/- 11,767.22)
```

Okay, _it became faster somehow_.

PR #72, opened 2024-10-18
Fix bug in mirgen for delay

tomoyanonymous:
This PR fixes the bug in the MIR generator for delay.

I don't know why but the condition of the function name was inverted.

### DIFF
--- a/mimium-lang/src/compiler/mirgen.rs
+++ b/mimium-lang/src/compiler/mirgen.rs
@@ -85,6 +85,8 @@ impl Context {
         };
         match max.to_expr() {
             Expr::Literal(Literal::Float(max)) => {
+                //need to evaluate args first before calculate state offset because the argument for time contains stateful function call.
+                let args = self.eval_args(&[*src, *time])?;
                 let max_time = max.parse::<f64>().unwrap();
                 let shift_size = max_time as u64 + DELAY_ADDITIONAL_OFFSET;
                 self.get_current_fn().state_sizes.push(StateSize {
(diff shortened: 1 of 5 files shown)


tomoyanonymous:
Found another bug. 

I added test cases but it causes sigabort.

```rust
fn counter(){
    self+1.0
}
fn dsp(){
    delay(10.0,counter(),5.0)
}
```
If I bound counter() with let, it passes the test.
```rust 
fn counter(){
    self+1.0
}
fn dsp(){
    let c = counter()
    delay(10.0,c,5.0)
}
```

In the MIR, statepos operation is not inserted if the call is done in the argument.

```
fn dsp []
upindexes:[]upper:0 state_size: 14(=[((number,number,number)->number*13),number])
  block 0
  reg(5)  := uint 1
  reg(6)  := call reg(5) [] ->number
  reg(7)  := float 5
  reg(8)  := delay 10 reg(6) reg(7)
  reg(9)  := ret reg(8) number
```

```
fn dsp []
upindexes:[]upper:0 state_size: 14(=[number,((number,number,number)->number*13)])
  block 0
  reg(7)  := alloc number
  reg(5)  := uint 1
  reg(6)  := call reg(5) [] ->number
  reg(8)  := store reg(7), reg(6), number
             pushstateidx 1(=[number])
  reg(9)  := load reg(7), number
  reg(10) := float 5
  reg(11) := delay 10 reg(9) reg(10)
             popstateidx  1(=[number])
  reg(12) := ret reg(11) number
```

tomoyanonymous:
Now both bugs are fixed.

PR #76, opened 2024-10-30
Midi Input Plugin

tomoyanonymous:
This PR implements experimental support for MIDI input devices.

MidiPlugin provides 2APIs: `set_midi_port("port_name")` and `bind_midi_note_mono(channel,default_note,default_velocity);`.

`bind_midi_note_mono` returns getter function of `(float,float)` value which is updated asynchronously by midi note event.(NoteOff is treated as NoteOn with 0 velocity).

Processing for raw MIDI events like midi plugin in VST cannot be realized for now.

(Note that MIDI device is not available for WSL. I tested only on macOS.)

```rust
let _ = set_midi_port("from Max 1")
fn osc(freq){
   ...
}
fn midi_to_hz(note){
    440.0*  (2.0 ^((note-69.0)/12.0))
}
let boundval = bind_midi_note_mono(0.0,69.0,127.0);
fn dsp(){
    let (note,vel) = boundval();

    let sig = note |> midi_to_hz |> osc 
    let r = sig * (vel /127.0);
    (r,r)
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,6 +116,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
+name = "atomic_float"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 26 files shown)


tomoyanonymous:
I think it's ready to merge for now though I found a bug that storing string type value in a global context other than first line causes crush. This bug seemingly caused by bytecode generator.

PR #81, opened 2024-11-03
Removed Nightly Feature

### LINKED ISSUE #16: Remove unstable feature to switch from a nightly channel to the stable
- `#![feature(box_patterns)]` if we use arena for recursive types, no longer needed(maybe [match_deref](https://crates.io/crates/match_deref) crate macro can be used for simple matching)
  - #12 
- `#![feature(iterator_try_collect)]` [`itertools`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.try_collect) can be used
- `#![feature(iter_collect_into)]` used only one location in `typing.rs`, no need to persist with this feature
- `#![feature(if_let_guard)]` no longer used in current code, can just remove
- `#![feature(lazy_cell)]` stabilized in v1.8.0, just remove it and set the minimum compiler version

tomoyanonymous:
close #16


tomoyanonymous:
I just realized that ‘test::bench‘ requires nightly channel. 
You need to run bench with ‘cargo +nightly bench‘. (Added in comments.)

PR #83, opened 2024-11-05
Remove or resolve some of my TODOs

yutannihilation:
I added several TODOs, but some of them are outdated and some are easy to resolve.

* `mimium-lang/src/ast_interpreter.rs`: `eval_extern()` is nested too deeply. I created a separate function.
* `mimium-lang/src/compiler/mirgen.rs`: This was simply my misunderstanding about `Tuple`. Just removed the TODO comment.
* `mimium-lang/src/compiler/parser/token.rs` & `mimium-lang/src/runtime/builtin_fn.rs`: As discussed in https://github.com/tomoyanonymous/mimium-rs/pull/31#issuecomment-2308142925, probably predefining symbols doesn't matter much. Just removed the TODO comments.
* `mimium-lang/src/types.rs`: Currently, `Type::apply_fn()` holds the actual logic, but it's used only via `TypeNodeId::apply_fn()`. So, I moved the code into `TypeNodeId` to reduce the round-trip conversions between these two types. Considering types are stored as `TypeNodeId` in most of the places, I guess this should cover most of the cases without`Type::apply_fn()` (and `Type::fold()` as well).

### DIFF
--- a/mimium-lang/src/ast_interpreter.rs
+++ b/mimium-lang/src/ast_interpreter.rs
@@ -159,91 +159,93 @@ fn eval_with_new_env(
     res
 }
 
+fn find_matched_builtin_fn(n: Symbol, tv: &[TypeNodeId]) -> Option<(Type, *const ())> {
+    builtin_fn::get_builtin_fns()
+        .iter()
+        .find_map(|(name, ty, ptr)| {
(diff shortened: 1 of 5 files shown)


tomoyanonymous:
Thanks as always! Looks okay.

yutannihilation:
Thanks for merging!

Issue #86, opened 2024-11-08
CSV backend seems broken


yutannihilation:
It seems it successfully finds the input device, but fails to pass it to `SystemPlugin`? In order to catch up with the recent updates, this might be a good practice for me (I'm yet to figure out how the plugins system works).

```
❯ cargo run --bin mimium-cli -- --output-format csv tmp.mmm

...

warning: `mimium-symphonia` (lib) generated 1 warning
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
     Running `target/debug/mimium-cli --output-format csv tmp.mmm`
[D] Filename: /home/yutani/repo/mimium-rs/tmp.mmm
[*] trying to connect default MIDI input device...
[*] Midi Input: Connected to Midi Through:Midi Through Port-0 14:0
ch0
thread 'main' panicked at mimium-midi/src/lib.rs:123:41:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

`mimium-midi/src/lib.rs:123:41` is around here.

https://github.com/tomoyanonymous/mimium-rs/blob/00a43378674b567a486a6ca5cc941597d34c4b31/mimium-midi/src/lib.rs#L121-L124

yutannihilation:
`after_main()` is called multiple times while `input` is `take()`n by the first call.

https://github.com/tomoyanonymous/mimium-rs/blob/00a43378674b567a486a6ca5cc941597d34c4b31/mimium-audiodriver/src/driver.rs#L104-L108

https://github.com/tomoyanonymous/mimium-rs/blob/00a43378674b567a486a6ca5cc941597d34c4b31/mimium-midi/src/lib.rs#L140

tomoyanonymous:
Does the panic happens on realtime audio? I guess the LocalBufferDriver somehow triggers `after_main` more than once.

tomoyanonymous:
https://github.com/tomoyanonymous/mimium-rs/blob/00a43378674b567a486a6ca5cc941597d34c4b31/mimium-audiodriver/src/backends/local_buffer.rs#L91 

This part was no longer needed.

BTW, in the current implementation, the GUI launches after emitting csv. GUI tools plugin should be turned off if the command line option is non-realtime context.

yutannihilation:
> Does the panic happens on realtime audio?

No, it works fine.

yutannihilation:
> This part was no longer needed.

[...]

PR #88, opened 2024-11-09
Do not launch GUI except for realtime audio mode

yutannihilation:
This pull request is based on this suggestion: https://github.com/tomoyanonymous/mimium-rs/issues/86#issuecomment-2463973245

* The GUI is launched only on realtime audio mode
* Users can use `--no-gui` to suppress GUI

As the logic around args became a bit complicated, this pull request pushes it into `RunOptions`, but I'm open to other suggestions.

### DIFF
--- a/mimium-audiodriver/src/backends/csv.rs
+++ b/mimium-audiodriver/src/backends/csv.rs
@@ -16,7 +16,14 @@ pub struct CsvDriver {
 }
 
 impl CsvDriver {
-    pub fn new(times: usize, csv_file: Box<dyn Write>) -> Self {
+    pub fn new<P: AsRef<Path>>(times: usize, path: &Option<P>) -> Self {
+        let csv_file: Box<dyn std::io::Write> = if let Some(path) = path {
+            let csv_file = File::create(path.as_ref()).unwrap();
+            Box::new(BufWriter::new(csv_file))
+        } else {
+            Box::new(BufWriter::new(std::io::stdout()))
+        };
+
         Self {
             driver: LocalBufferDriver::new(times),
             csv_file,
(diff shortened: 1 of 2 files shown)


tomoyanonymous:
Thanks! `RunMode` option looks nice.
I'm wondering if I should add the machanism so that user can turn on and off plugin explicitly from the CLI option, but for now it is okay.

yutannihilation:
> turn on and off plugin explicitly from the CLI option

Sounds good. I came up with a naive idea of adding an associated item like `const LABEL: &str` to `Plugin` trait to match with the CLI option (e.g. `--enable-plugin LABEL`). But, it seems this doesn't work because this would make it impossible to create a trait object of `Plugin`. I think we need some research about how to implement such a system.

```
error[E0038]: the trait `plugin::Plugin` cannot be made into an object
  --> mimium-lang\src\plugin.rs:72:40
   |
72 | pub fn get_extfun_types(plugins: &[Box<dyn Plugin>]) -> impl Iterator<Item = ExtFunTypeInfo> + '_ {
   |                                        ^^^^^^^^^^ `plugin::Plugin` cannot be made into an object
   |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
  --> mimium-lang\src\plugin.rs:21:11
   |
20 | pub trait Plugin {
   |           ------ this trait cannot be made into an object...
21 |     const LABEL: &str;
   |           ^^^^^ ...because it contains this associated `const`
   = help: consider moving `LABEL` to another trait
```

tomoyanonymous:
I guess it can be object-safe if we change the value from an associated constant like ` const LABEL: &str;` to the method that returns `&str`(or just `String`)

yutannihilation:
Ah, thanks for the hint. Should I address it in this pull request as well?

tomoyanonymous:
Not necessary! Then I'll merge this for now.

yutannihilation:
Thanks! Maybe I'll try if I can implement the option in that way when I have time.

PR #90, opened 2024-11-09
Calculate output buffer size based on `dsp_ochannels`

### LINKED ISSUE #89: audiodriver crashes when the number of output channels doesn't match
This crashes with the current `dev` branch, but I think this used to work without errors.
Is this a known limitation of the current implementation?

```
fn dsp(){
  0.0
}
```

```
[*] output device 1 - ... channels: 2
[*] in:true out:true
Press Enter to exit
thread 'cpal_wasapi_out' panicked at mimium-audiodriver\src\backends\cpal.rs:77:57:
range end index 8192 out of range for slice of length 4096
```

yutannihilation:
Fix #89 

As suggested in [this comment](https://github.com/tomoyanonymous/mimium-rs/issues/89#issuecomment-2466079264), current implementation doesn't refer to `dsp_ichannels` and assume it is fixedly `2`. 

> Size of the dst slice in process function varies depending on the hardware output channels, however, the size of the localbuffer has not changed on the initialization correctly...

The problem is that, while the actual number of channels is determined by the combination of `dsp()` and the hardware. So, the calculation of the buffer size needs to match with this logic.

https://github.com/tomoyanonymous/mimium-rs/blob/00a43378674b567a486a6ca5cc941597d34c4b31/mimium-audiodriver/src/backends/cpal.rs#L89-L108

### DIFF
--- a/mimium-audiodriver/src/backends/cpal.rs
+++ b/mimium-audiodriver/src/backends/cpal.rs
@@ -224,7 +224,6 @@ impl Driver for NativeDriver {
     fn init(&mut self, ctx: ExecContext, sample_rate: Option<SampleRate>) -> bool {
         let host = cpal::default_host();
         let dsp_ichannels = 1; //todo
-        println!("{}", self.buffer_size_per_ch);
         let dsp_ibuffer_size = self.buffer_size_per_ch * dsp_ichannels;
         let (prod, cons) = HeapRb::<Self::Sample>::new(BUFFER_RATIO * dsp_ibuffer_size).split();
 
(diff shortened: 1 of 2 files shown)


yutannihilation:
While the code is a bit messy, I think this is ready for review now. This pull request fixes the logic of output buffer, but gave up that of input buffer. That's a battle for another day...

tomoyanonymous:
I found that there's some audio glitch with specific situations when I use with Blackhole(16ch virtual audio driver), and crashes when I use my bluetooth headphone(it gives a non-fixed sized buffer regardless the buffersize is `Fixed` in cpal option.) I will fix them in succeeding PR!

PR #91, opened 2024-11-12
Fixed the crush and glitch in multiple output channels

tomoyanonymous:
Continue from #90.

I removed `adjusted_ochannels` in the end and it look working without it.

Could you try if there's no probelm? > @yutannihilation

### DIFF
--- a/mimium-audiodriver/src/backends/cpal.rs
+++ b/mimium-audiodriver/src/backends/cpal.rs
@@ -48,10 +48,6 @@ pub fn native_driver(buffer_size: usize) -> Box<dyn Driver<Sample = f64>> {
 struct NativeAudioData {
     pub vmdata: RuntimeData,
     dsp_ochannels: usize,
-    // The number of output channels might differ between dsp() and the actual
-    // hardware. Depending on the combination, some channles might be duplicated
-    // or dropped.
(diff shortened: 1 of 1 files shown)


yutannihilation:
Thanks, confirmed! This looks simpler and better than mine!

mimium-audiodriver/src/backends/cpal.rs
```diff
@@ -111,15 +98,15 @@ impl NativeAudioData {
                     for i in 0..dspout {
                         o[i] = res[i] as f32;
                     }
-                    for i in dspout..hout {
-                        o[i] = 0.;
-                    }
                 }
                 (1, 2) => {
                     o[0] = res[0] as f32;
                 }
```
yutannihilation, line 104: This case should probably be covered by the next `(hout, _)` case.

tomoyanonymous: Indeed. I was wondering if we should mix stereo signal when the case of mono output hardware && stereo DSP code. For now, I'll simply delete the case above.

yutannihilation: I see. Either should be fine.

PR #98, opened 2024-11-14
Bump cargo-dist and settings for publishing in crate.io

tomoyanonymous:
Bumps cargo-dist to 0.25.

It looks that cargo-dist now uses `dist-workspace.toml` instead writing everything into `Cargo.toml`.

Also, I removed the line like `{path = "../mimium-lang" }` for in-workspace dependencies and uses `{workspace = true}` because `cargo publish` allows only explicitly versioned dependencies.

I refered to the [`egui`'s Cargo.toml](https://github.com/emilk/egui/blob/master/Cargo.toml).


tomoyanonymous:
I'm wondering how can I include `mimium-lang/lib/**.mmm` in the release bundle.
`cargo-dist` has `include` option but it does not allow glob pattern.

https://opensource.axo.dev/cargo-dist/book/reference/config.html#include

Is it possible to include such asset with `build.rs`? If the people who installed mimium-cli with `cargo install` can not get such assets with the `cargo-dist` dependent bundling in the first place...

tomoyanonymous:
Test publish:
https://crates.io/crates/mimium-lang

yutannihilation:
> Is it possible to include such asset with `build.rs`? In the first place, the people who installed mimium-cli with `cargo install` can not get such assets with the `cargo-dist` dependent bundling...

Guessing from https://github.com/axodotdev/cargo-dist/issues/193#issuecomment-1699096293, it seems the recommendation is to include the whole directory instead of a glob.

> Is it possible to include such asset with `build.rs`?

I think this is possible, but it might not be very obvious where such files should be installed.

tomoyanonymous:
Hmm, it looks impossible to include assets for an application provided through `cargo install`. I'll take care only cargo-dist

https://github.com/rust-lang/cargo/issues/11683#issuecomment-1419190756

tomoyanonymous:
https://crates.io/users/tomoyanonymous

Finished test release on crate.io. 
I could not use cargo release because there is limit for publishing multiple new crates at the same time.
I'll test bumping a version with cargo-release later.

PR #103, opened 2024-11-25
Remove unused `adjusted_ichannels`

yutannihilation:
I added `adjusted_ochannels` and `adjusted_ichannels` in #90, but later it turned out these are unnecessary, so `adjusted_ochannels` was removed in #91. It's still unclear what the implementation for input will be, but I bet `adjusted_ichannels` can be removed for now for consistency.


tomoyanonymous:
Thanks. I think it looks OK. 

Passing input signal to dsp code is the next task. Maybe similar variables will be needed again when the input and output channels are different.

PR #110, opened 2024-12-16
Fix incorrect type inference

tomoyanonymous:
This PR fixes the incorrect type unification in function type.

Previously, the incorrect higher-order funtion like the code below  was not treated as an error.

```rust
let doublethunk = | | {| | 1.0} 

fn hof_ident(f:()->float)-> ()->float {
    f
}

let f = hof_ident(doublethunk) //this should be an error

fn dsp()->float{
 f()
}
```

This PR also fixes incorrect type definition of primitive log function.

### DIFF
--- a/mimium-cli/examples/multiosc.mmm
+++ b/mimium-cli/examples/multiosc.mmm
@@ -6,10 +6,10 @@ fn osc(freq){
 fn amosc(freq,rate){
   osc(freq + osc(rate)*4000.0)
 }
-fn replicate(n,gen:()->(float,float)->float){
+fn replicate(n,gen){
     if (n>0.0){
         let c = replicate(n - 1.0,gen)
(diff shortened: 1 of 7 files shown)


tomoyanonymous (commented):
This PR also fixes instantiation of type scheme variable.
Now this kind of recursive higher-order function works without type annotation.

```rust
//this is exactly same as hof_state.mmm but now works without explicit type annotation.
fn counter(inc){
    self+inc
}

fn replicate(n,gen){
    if (n>0.0){
        let c = replicate(n - 1.0,gen)
        let g = gen()
        |x| {g(x) * n + c(x)}
    }else{
        |x| { 0.0 }
    }
}
let mycounter = replicate(3.0,| |counter);
fn dsp(){
    mycounter(0.1)
}
```

PR #117, opened 2024-12-23
fixed type check  of block

tomoyanonymous:
This PR fixes the bug in the type checker so that the  block with curly brace creates local variable scope correctly.

```rust
//Curly brace creates local scope. should be 3.0
fn dsp(){
    let test = 0
    {
        let local = 1.0
    }
    let local = 3.0
    test + local
}
```

```rust
//Curly brace creates local scope. This emits error.
fn dsp(){
    let test = 0
    {
        let local1 = 1.0
    }
    test + local1 // local1 cannot be found from this scope.
}
```


tomoyanonymous:
I'll merge this though the logic of variable finding in mirgen looks little bit different from it of type-checker but working somehow...should be addressed in the succeeding refactoring of mirgen.

PR #124, opened 2025-06-19
Array implmentation

### LINKED ISSUE #123: Implement: basic array syntax
Add an array creation/access functionality.

The array type proposed here is:

- at the type level, the length are variable(2 vector with different lengths have a same type)
- but they are basically immutable and have no method of push/pop in the value level
- in the VM, they are allocated on special space in heap like closure storage, thus have single word size. 
  - The internal memory layout is just a "fat pointer", meaning it is single word of the length & actual data space.

This spec is basically same as it implemented in v0.4.

Syntax is like below:

```
let myarr = [1.0, 2.0, 3.0];
let elem = myarr[0]
```

The syntax for access to the element `myarr[0]` is syntactic sugar to `array_access_interp(idx:float)`.
The access in an index between integer is automatically linear-interpolated, and access out of bound returns 0.
So this is inefficient in the case of specific usecase that the access is assured to never be out of bound but we allow it for the temporary spec.

They have special method `length_array()` to obtain a length of the array. This method is essentially a generic method but can be implemented as single method because the number of elements is always stored in the head of fat-pointer.

- [x] implement parser
  - [x] implement test for parser
- [x] implement type checker
- [x] implement MIR 
- [x] implement VM instruction bytecode spec
- [ ] implement VM internal working
  - [x] implement `length_array()` 
  - [ ] ~`array_access_interp()`~ (don't do because the behavior for array of aggregate types are unclear)
- [x] implement integrated test for array

tomoyanonymous:
Close #123 
This pull request introduces several updates across the `mimium-rs` project, focusing on enhancing the documentation in the Memory Bank and adding support for array literals and array types in the language implementation. These changes improve both the project's organization and its functionality.

* [`mimium-lang/src/ast.rs`](diffhunk://#diff-1de172e69596ec99c6994f4d40db78aa1e1ad2ee9cd511e45a33d29ef40d2f1aR45): Added support for array literals in the Abstract Syntax Tree (AST) by introducing the `ArrayLiteral` variant in the `Expr` enum and its corresponding `MiniPrint` implementation. [[1]](diffhunk://#diff-1de172e69596ec99c6994f4d40db78aa1e1ad2ee9cd511e45a33d29ef40d2f1aR45) [[2]](diffhunk://#diff-1de172e69596ec99c6994f4d40db78aa1e1ad2ee9cd511e45a33d29ef40d2f1aR129-R132)
* [`mimium-lang/src/compiler/parser.rs`](diffhunk://#diff-67617bce20c776e4bdeafd12c5c03da850ac8a91b88c8b1b2cb9c14d0086ac28R52-R63): Extended the parser to support array types, enabling syntax like `[T]` for arrays.


tomoyanonymous (commented):
I guess it's now ready for merge for the initial implementation.

Current spec:

- Arrays are immutable in the value level(no push/pop method)
- However, type information has no information of array length(variable length in the type level)
- Not only the array of number, array of aggregate types such as tuple can be used.
- In the previous implementation, the array had automatic interpolation & boundary check feature for decimal index access but currently not implemented. Reason for this is because the behavior for array of aggregate type is unclear.
- It's memory leaky. It's allocated on the heap but no garbage collection for now. So do not create new array in dsp context. It may be resolved by introducing new instruction like "droparray"

tomoyanonymous (commented):
I guess it's now ready for merge for the initial implementation.

Current spec:

- Arrays are immutable in the value level(no push/pop method)
- However, type information has no information of array length(variable length in the type level)
- Not only the array of number, array of aggregate types such as tuple can be used.
- In the previous implementation, the array had automatic interpolation & boundary check feature for decimal index access but currently not implemented. Reason for this is because the behavior for array of aggregate type is unclear.
- It's memory leaky. It's allocated on the heap but no garbage collection for now. So do not create new array in dsp context. It may be resolved by introducing new instruction like "droparray"

PR #128, opened 2025-07-02
Implementing Basic Record Syntax

### LINKED ISSUE #127: Create basic record syntax and type inference system

tomoyanonymous:
close #127

- [x]  implement parser
- [x] implment patern capture syntax parser for record type
- [x] implement type inference system
- [x] implement MIR generation 

maybe in future PR?

- bugfix of record types w/ different order of keys
- structural subtyping implementation

### DIFF
--- a/mimium-lang/src/compiler/parser.rs
+++ b/mimium-lang/src/compiler/parser.rs
@@ -289,6 +289,7 @@ where
     let ctx = ctx.clone();
     let dot = apply
         .then(just(Token::Dot).ignore_then(dot_field()).repeated())
+        .boxed() // this boxing is necessary for windows CI environment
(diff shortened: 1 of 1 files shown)


tomoyanonymous:
Almost ready but it still contains one severe bug.
If the 2 record type which have same keys but different order, the index for "GetElement" in MIR generation may be wrong.

I'm wondering if I should fix within this PR or not, because it will be fixed when the structural subtyping is introduced in the next PR anyway

```rust
let a = {keya:100.0, keyb:(200.0,300.0)}
let b = {keyb:(0.0,0.0), keya: 0.0}
b = a

fn dsp(){
    b.keya //should be 100.0
}
```

PR #129, opened 2025-07-03
Windows CI test

tomoyanonymous:
Test for Windows CI somehow fails in #128. Rolled back to the next commit to the last succeeded test

### DIFF
--- a/.cline/memory-bank/activeContext.md
+++ b/.cline/memory-bank/activeContext.md
@@ -2,47 +2,163 @@
 
 ## Current Work Focus
 
-The current focus is on implementing a Language Server Protocol (LSP) server for the mimium programming language. This will provide IDE features such as syntax error reporting, code completion, and hover information to improve the developer experience.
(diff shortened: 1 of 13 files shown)


yutannihilation:
FWIW, it seems these failures are not reproducible on my Windows machine.

tomoyanonymous:
This error is not just a linker error but Stack buffer overrun actually. The issue is raised that the error messages is not user-friendly(https://github.com/rust-lang/rust/issues/141757)

And I suspect this error is caused by that the parser-combinator generates too-much complex type signature and it causes crush on some lightweight environment. We need to add `.boxed()` to reduce type complexity.

yutannihilation:
Ah, thanks for the explanation. It sounds more complex than I expected...

tomoyanonymous:
https://github.com/mimium-org/mimium-rs/pull/128/commits/00ffda8998d5408be99ab803200470736bc5dff4 

Fixed in this commit in the original branch. I'll close this PR.

Issue #135, opened 2025-07-14
Implementing Hygienic Macro(multi-stage computation)


tomoyanonymous:
## Overview

Implementing macro is super important for mimium.

Because audio processing generally treats 2 steps of computation, building computation graph at first, and running it.

If we want to parameterize the audio graph, currently, we need to "instantiate" the graph, using higher-order function like this.

```rust
fn bandpass(x,freq){
      //...
    }
fn filterbank(n,filter_factory:()->(float,float)->float){
  if (n>0){
    let filter = filter_factory() 
    let next = filterbank(n-1,filter_factory)
    |x,freq| filter(x,freq+n*100)
             + next(x,freq)
  }else{
    |x,freq| 0
  }
}
let myfilter = filterbank(3,| | bandpass)
fn dsp(){
      myfilter(x,1000)
}
```

This is cumbersome compared to faust, which can write complex parametric audio graphs using `par`, `seq` or even term-rewriting macro with pattern matching. However, faust's macro is sometimes problematic because the macro has an independent semantics, different from block-diagram algebra, so the macro can generate invalid codes and fails to compile(and usually it is difficult to know the reason of the failure from the error messages). 

What I propose for hygenic macro of mimium, is "Multi-stage computation", which is used in BER MetaOCaml and recent Scala (after v3).
Multi-stage computation is one of the extended lambda-calculus, to describe macro as also a normal lambda calculus.

It adds 2 language primitives "quote/bracket" "unquote/splice/escape". The former describes the expression will be evaluated in next stage, and the latter describes the expression will be embedded in current stage with expanding the quoted expression inside. These are similar to LISP's "quote" and "quasiquote". Multi-stage computation is something equivalent to it in the statically-typed language world.

Type system will have an additional "code" type constructor to distinct to which stage the expression will be evaluated.

Overall, the macro system with multi-stage computation is nice because all type check/inference is done before expanding macro, which will improve readability of the errors.

## Syntax

- The "quote" syntax will be, the ` `expr`, prefix with the single backquote.
- The "splice" syntax will be, the `$expr`, prefix with the dollar-sign(this is equivalent to the Scala).
  - the large expressions can be wrapped either with `()` `{}`, without changing current syntax.
- Type annotation for "code" type will be ``type` like quote syntax.

The filterbank example can be written with these syntax like this.

```rust
fn filterbank(n,filter:`(float,float)->float)->`(float,float)->float{
  `{ if (n>0){
[...]

PR #136, opened 2025-07-14
Implement multi-stage computation intepreter 

tomoyanonymous:
basic implementation of #135

### DIFF
--- a/mimium-lang/src/compiler/bytecodegen.rs
+++ b/mimium-lang/src/compiler/bytecodegen.rs
@@ -804,7 +804,7 @@ impl ByteCodeGenerator {
             .collect();
         self.program.file_path = mir.file_path;
         self.program.iochannels = mir.get_dsp_iochannels();
-        // log::info!("iochannels: {:?}", self.program.iochannels.unwrap());
+        log::debug!("iochannels: {:?}", self.program.iochannels.unwrap());
         let _io = self.program.iochannels.unwrap();
         self.program.clone()
     }
--- a/mimium-lang/src/compiler/mirgen.rs
+++ b/mimium-lang/src/compiler/mirgen.rs
@@ -593,20 +593,16 @@ impl Context {
                 (reg, Type::Array(elem_ty).into_id())
             }
             Expr::ArrayAccess(array, index) => {
-                let (array_v, array_ty) = self.eval_expr(*array);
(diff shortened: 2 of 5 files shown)


tomoyanonymous:
Now it looks working.

Currently, if the last expression of main program is `Code` type, like the example above, macro expansion will be enabled. All other codes outside the bracket with backquote, are regarded as level-0(macro) code. 
Here, function definition with `fn name()` is only allowed in toplevel, not allowed inside curly braces.

So I'm going to add some handy syntactic sugar to distinct macro section and others, without wrapping everything with brackets. Somthing like this.

```rust
//!macro

fn genpower(n:float){
    letrec aux = |n1:float,x| {
        if (n1>0){
           `{$( aux(n1-1,x)) * $x }
        }else{
           `1
        }
    }
    `{|x:float| $(aux(n,`x)) }
}

//!


fn dsp(){
 2 |> $(genpower(5)) 
}

```

tomoyanonymous:
Added example of `cascadeosc` with macro evaluation. Currently the curly braces for stages are deeply nested so hard to read, but it works.

```rust
//same as cascadeosc.mmm, but written with new multi-stage computation feature.
// include("osc.mmm")
`{
let PI = 3.14159265359

let phasor_shift =  |freq,phase_shift|{
    (self + freq/samplerate + phase_shift)%1.0
}

let sinwave = |freq,phase|{
    phasor_shift(freq,phase)*2.0*PI |> sin
}
let osc = |freq|{
  sinwave(freq,0.0)
}
${
    letrec cascade = |n,gen|{
        if (n>0.0){
            let c = cascade(n - 1.0 ,gen)
            let multiplier = 1.0-(1.0/(n*2.0)) |> lift_f
            `{|rate| rate + ($gen)(rate/2.0)* 0.5 * rate* $multiplier  |> $c }
        }else{
            `{|rate| ($gen)(rate)}
        }
    }
`{
let dsp = | | {
    let r =  200 |> $(cascade(20.0,`osc))
    (r,r)
    }
}
    }
}
```

Benchmark result for cascading 40 oscillators, with and without macro was slightly better when macro was used (about 5%).

```
test tests::multistage::bench_multistage        ... bench:     931,200.00 ns/iter (+/- 26,897.85)
test tests::multistage::bench_multistage_normal ... bench:     977,161.97 ns/iter (+/- 28,977.21)
```

PR #159, opened 2025-09-30
”Live” state updating feature over different version of codes

tomoyanonymous:
This pull request introduces a new internal state management system between different version of code, which enables "almost live" coding in mimium.

This mechanism is achieved through the following steps. First, as a key feature of mimium, for functions where the call flow is determined at compile time (i.e., those that do not dynamically alter the signal processing flow, such as higher-order functions), the internal state used for delay and feedback is statically allocated in a single contiguous memory region.
If only specific parameters like delay time change without altering the call tree, this linear memory region can be reused directly in the new VM. This allows the new code to start executing without interrupting the delay or reverb tail.

Furthermore, if we can detect which parts of the call tree were inserted, deleted, or replaced, even if the call tree structure changes, we can inherit the previous internal state with minimal data copying from the old linear memory to the new linear memory.

This mechanism is abstracted within the newly added `state_tree` crate.

Under the current mechanism, internal state dependent on external I/O, such as the value of a GUI slider, gets reset during recompilation. In the future, this might be resolved by introducing a new primitive that allows initializing internal state with arbitrary values and persisting that state during recompilation.

Alongside the live VM update mechanism, several disruptive changes are planned for the plugin API to asynchronously send new VMs to the audio thread. This includes some temporary workaround code because interned values (like TypeNodeId or Symbol) cannot be reused across threads.


tomoyanonymous:
Made large changes in SystemPlugin API and more.

Especially, I changed `SESSION_GLOBALS` from thread_local Rc<RefCell<>> to LazyLock<Mutex<>>.

This is for making compiler to be asynchronous on the other threads, but still some TypeNodeId or Symbol were used in the runtime, and it was difficult to make it work without this workaround.

Also, TypeVar is now use `Arc<Mutex<>>` instead of `Rc<RefCell<>>`. 

I marked `compiler::Context`  as `Send` but not sure if it is okay because some macro function modifies system plugins' internal states during recompilation.
In fact, some random crash happens on recompilation now. I'm not sure if it is exactly caused by that part, need to ivestigate more.

tomoyanonymous:
I'm gonna merge this while leaving this weird bug in the separated issue... The behavior implies some unsafe code around internal state management collide with upvalue management.

PR #170, opened 2025-10-15
fix crash when accessing self keyword in nested block scope

karnpapon:
**Description**

This issue occurred when I tried to run `example/0b5vr.mmm` and encountered a crash with `an invalid memory reference` error. After investigating, I found that the problem potentially stems from functions using the `self` keyword being called from within nested block scopes (in this case, from the `integ` function in `example/0b5vr.mmm`). This happened because the VM wasn't properly allocating global state storage for functions requiring state in these contexts.

**Solution:**
Added state size validation in `Instruction::Call` to ensure the global state buffer is adequately sized before function execution.

I also added a unit test `test_self_nested_block_allocation` to verify the fix (though I'm not sure I implemented it correctly, as bytecodes are new to me 😵).

### DIFF
--- a/crates/lib/mimium-lang/src/runtime/vm.rs
+++ b/crates/lib/mimium-lang/src/runtime/vm.rs
@@ -771,6 +771,16 @@ impl Machine {
                 }
                 Instruction::Call(func, nargs, nret_req) => {
                     let pos_of_f = Self::get_as::<usize>(self.get_stack(func as i64));
+                    let func_proto = self.get_fnproto(pos_of_f);
+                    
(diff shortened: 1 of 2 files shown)


tomoyanonymous:
Thank you for investigating again! It seems that the crush is caused by some inappropriate code generation by mirgen or bytecode generator. Statestorage will never be resized while execution.
I think the conditional branch with stateful function still causes some bugs, maybe related to that.

I'll take a look further later.

karnpapon:
ah, got it! gave it a try and it worked perfectly. thanks Tomoya 🙏

PR #180, opened 2026-02-03
Feature: Module Syntax& Semantics 

tomoyanonymous:
This PR adds module syntax & semantics to mimium.

The feature is heavily based on it of Rust. Module is not a first-class and used only for separating namespace.

- module can be created with `mod { statements }`.
- External file module can be created with `mod name`. `./name.mmm` will be the content of the module
- Users can control visibility with `pub` prefix as same as Rust.
- Users can import declarations defined in module with `use name`. Multiple imports with curly braces and wildcard are supported.
- Different from Rust, semicolons in the line end is not required.
- Semantic highlight in a language server for module declaration is also supported in this PR.
- Macro declaration (`#stage(macro)`) is enclosed within each modules. Each module implicitly starts with stage main.

```rust
mod math {
    pub fn double(x) { x * 2.0 }
    pub fn triple(x) { x * 3.0 }
}
use math::*

fn dsp() {
    double(10.0) + triple(10.0)
}
```

```rust
mod outer {
    mod inner {
        pub fn secret() {
            42.0
        }
    }
    pub fn exposed() {
        // Using relative path: inner::secret() instead of outer::inner::secret()
        inner::secret()
    }
}

fn dsp() {
    outer::exposed()
}
```

```rust
mod test{
    #stage(macro)
    pub fn macrofn1(x){
        `{$x+$x}
    }
}

fn dsp(){
    test::macrofn1!(`1)
}
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,4 +45,4 @@ jobs:
             target/
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
-      - run: cargo test
+      - run: cargo test -- --test-threads=1


tomoyanonymous:
@copilot LanguageServerのDiagnosticsでモジュール外からモジュール内の変数参照がVariable not found のエラーになってしまうので原因を特定して

